PR #42, opened 2018-03-21
Bugfix/adaptations for internal modules

JonasHallFnx:
Changes needed for us to use the OSS modules internally.

### DIFF
--- a/config/pom.xml
+++ b/config/pom.xml
@@ -36,4 +36,20 @@
 		</dependency>
 	</dependencies>
 
+	<build>
+		<plugins>
+			<plugin>
+				<groupId>org.apache.maven.plugins</groupId>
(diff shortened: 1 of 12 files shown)


config/pom.xml
```diff
@@ -36,4 +36,20 @@
...
+	<build>
+		<plugins>
+			<plugin>
+				<groupId>org.apache.maven.plugins</groupId>
+				<artifactId>maven-jar-plugin</artifactId>
+				<version>3.0.2</version>
+				<executions>
+					<execution>
+						<goals>
+							<goal>test-jar</goal>
```
magnusp: Nice! This solves the problem of not having the MockAppender available?

JonasHallFnx: it would, yes

Issue #13, opened 2017-12-01
Deploy to maven central


JonasHallFnx:

splitfeed:
Done!

http://search.maven.org/#search%7Cga%7C1%7Cse.fortnox.reactivewizard

PR #43, opened 2018-03-22
Setup for deployment to Maven Central

splitfeed:
First step to resolving #13


JonasHallFnx:
Do we also need to add the test-jar to the deployment? We have test-jar in jaxrs and config modules.

PR #49, opened 2018-03-27
Add support for decorating your response with response headers

JonasHallFnx:
The headers are put inside a subclass of the Observable, which means that this decoration must be the last thing you do be fore returning the result from your resource method.

### DIFF
--- a/jaxrs/src/main/java/se/fortnox/reactivewizard/jaxrs/response/ResponseHeaders.java
+++ b/jaxrs/src/main/java/se/fortnox/reactivewizard/jaxrs/response/ResponseHeaders.java
@@ -1,0 +1,46 @@
+package se.fortnox.reactivewizard.jaxrs.response;
+
+import rx.Observable;
+import se.fortnox.reactivewizard.jaxrs.response.JaxRsResult;
+
+import java.util.Map;
+
+/**
+ * Allows you to return response headers together with your result.
+ */
+public class ResponseHeaders {
+
+    /**
+     * Use this to wrap your Observable with some headers. This must be the last decoration before returning from your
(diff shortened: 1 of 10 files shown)


jaxrs/src/main/java/se/fortnox/reactivewizard/jaxrs/response/ResponseHeaders.java
```diff
@@ -0,0 +1,82 @@
...
+import rx.Observable;
+import rx.Observer;
+
+import java.util.Map;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+/**
+ * Allows you to return response headers together with your result.
+ */
+public class ResponseHeaders {
```
splitfeed: The name here indicates that it is the headers of a response, while it's actually a response that also contains headers. Could it be named simply Response or ResponseWithHeaders if something more verbose is wanted?

JonasHallFnx: The class is actually not holding the response, that is done by the private class ObservableWithHeaders.

The ResoponseHeaders class actually just has the withHeaders method, to do the decoration.

I am open to other names, but ResponseWithHeaders would be misleading.

tsfcode: So the class is more of a ResponseDecorator?

JonasHallFnx: @tsfcode you nailed it!

PR #51, opened 2018-04-06
Load log4j.properties from multiple locations

JonasHallFnx:
Needed becuase we have configuration in different modules.
Also changed so that you can configure appenders fully from yml without specifying anything in the log4j.properties.


splitfeed:
Would it be possible to add a test to verify that it actually reads multiple sources correctly?

JonasHallFnx:
@splitfeed I think testing this will be hard, because having multiple files with the same name on the classpath is tricky in a test. I could refactor the code to use dependency injection and inject something that gives me the Enumeration<URL> of resources, but this feature is only used when running in intelliJ, not when using the fatjar.

splitfeed:
Fair enough :)

PR #52, opened 2018-04-09
Add tests for deserializers

splitfeed:


JonasHallFnx:
Looks good! Perhaps you can remove some tests from JaxRsResourceTest that tests the same deserialization?

splitfeed:
@JonasHallFnx After spending a while looking through JaxRsResourceTest, I'm not sure I could remove anything there without adding more tests for other classes. It seems each of those tests seem to test a bunch of other aspects of deserialization as well, like the DeserializationFactory and that it in turn is used in all the right places.

Are there any particular tests that seem redundant to you? I feel that some might make better sense as separate unit tests, but that'll have to wait for another day :)

PR #53, opened 2018-04-09
ReflectionUtil - improve member lookup

erljef:
This change allows for better field- and method lookup when using class hierarchies.

The two issues this fixes are 
1) Allowing ReflectionUtil to create Getter and Setter instances for non-visible fields and methods in a superclass.
2) Allowing Getters and Setters to properly determine the return types and parameter types of fields and methods in a generic superclass.

### DIFF
--- a/utils/src/main/java/se/fortnox/reactivewizard/util/AccessorUtil.java
+++ b/utils/src/main/java/se/fortnox/reactivewizard/util/AccessorUtil.java
@@ -1,0 +1,30 @@
+package se.fortnox.reactivewizard.util;
+
+import sun.reflect.generics.reflectiveObjects.ParameterizedTypeImpl;
+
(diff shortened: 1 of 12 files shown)


magnusp:
Trivial impact on performance?

erljef:
@magnusp Good point, will do some tests.

@JonasHallFnx Sounds like a good idea. I have no experience with using it but from what I can see it is way faster than using Method.invoke() (close to precompiled code) and seems to be the preferred way of building serialization/deserialization.

erljef:
The performance is worse when looking up getters and setters for superclass methods (due to having to search the class hierarchy for the members). However, that lookup is done once during setup (creation of serializers/deserializers) and the runtime performance of calling invoke (which we did before as well) is the same.

PR #54, opened 2018-04-13
Fix issues with subclassing non-generic classes.

erljef:
With my changes to better support inheritance with our reflection stuff I failed to take into account that you can, in fact, use classes without generics... This PR fixes that.


erljef:
Hmmm, I'm running into some issues (somewhat related to the "missing branches" in AccessorUtil) that I'll have to resolve before this can be merged.

PR #56, opened 2018-04-17
Move DAO to open source project

splitfeed:


dao/src/main/java/se/fortnox/reactivewizard/binding/scanners/DaoClassScanner.java
```diff
@@ -0,0 +1,25 @@
...
+import se.fortnox.reactivewizard.db.Update;
+
+import javax.inject.Singleton;
+import java.lang.reflect.Executable;
+
+@Singleton
+public class DaoClassScanner extends AbstractClassScanner {
+    @Override
+    public void visit(FastClasspathScanner fastClasspathScanner) {
+        fastClasspathScanner.matchSubinterfacesOf(Dao.class, this::add);
```
JonasHallFnx: This can be removed

dao/src/main/java/se/fortnox/reactivewizard/db/Dao.java
```diff
@@ -0,0 +1,5 @@
+package se.fortnox.reactivewizard.db;
+
+public interface Dao {
```
JonasHallFnx: This can be dropped

dao/src/main/java/se/fortnox/reactivewizard/db/Named.java
```diff
@@ -0,0 +1,12 @@
+package se.fortnox.reactivewizard.db;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.PARAMETER)
+public @interface Named {
```
JonasHallFnx: Perhaps we should remove this and use javax.binding.Named, or simply just require that you use compiler option "-paramaters"

dao/src/main/java/se/fortnox/reactivewizard/db/Schema.java
```diff
@@ -0,0 +1,12 @@
+package se.fortnox.reactivewizard.db;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.PARAMETER)
+public @interface Schema {
```
JonasHallFnx: We should drop this as well I think

dao/src/main/java/se/fortnox/reactivewizard/util/JsonUtil.java
```diff
@@ -0,0 +1,55 @@
...
+import com.fasterxml.jackson.databind.DeserializationFeature;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.ObjectReader;
+import com.fasterxml.jackson.databind.type.TypeFactory;
+import com.fasterxml.jackson.databind.util.StdDateFormat;
+
+import java.lang.reflect.Type;
+import java.util.Map;
+
+public class JsonUtil {
```
JonasHallFnx: This should be dropped and our JsonDeserializerFactory should be used.

dao/src/main/java/se/fortnox/reactivewizard/db/Schema.java
splitfeed: I actually did remove that, but borked my workspace and lost the change. Thanks for catching it!

dao/src/main/java/se/fortnox/reactivewizard/db/Named.java
splitfeed: Are there any use cases for @Named that we lose by dropping it and going with -parameters?

splitfeed: Code-wise it looks like we'd get rid of a few lines of extra logic and if everyone can use -parameters they will get cleaner DAO:s as well.

dao/src/main/java/se/fortnox/reactivewizard/util/JsonUtil.java
splitfeed: I've managed to get rid of all but the convertValue method, is there any existing implementation that replaces that one?

JonasHallFnx: Just @Inject an ObjectMapper and use that.

dao/src/main/java/se/fortnox/reactivewizard/db/Named.java
JonasHallFnx: Funny that you accidentally referred to a github user there ;-) Maybe we get our first outsider!

I don't think we loose anything, but I believe there might be a migration for those who have a conflict between the value in the annotation and the name of the parameter. But if the annotation is removed, they will get a compilation error so that should be obvious.

Yolo!

erljef: We would lose some flexibility and force all projects to use -parameters. But I think it's acceptable to remove it - we can always add it again should the need arise.

dao/src/main/java/se/fortnox/reactivewizard/db/query/parts/ParamQueryPart.java
```diff
@@ -0,0 +1,134 @@
...
+        Optional<PropertyResolver> propertyResolver = argResolver.subPath(subPath);
+        if (!propertyResolver.isPresent()) {
+            throw new RuntimeException("Properties " + subPath + " cannot be found in " + argResolver);
+        }
+        return new ParamQueryPart(argIndex, propertyResolver.get());
+    }
+
+    protected PreparedStatementParamSetter createParamSetter(Type type) throws SQLException {
+        Class<?> rawType = ReflectionUtil.getRawType(type);
+        if (java.util.Date.class.isAssignableFrom(rawType)) {
```
JonasHallFnx: This is the only usage of java.util.Date, which required the dates module to be migrated as well. I vote for removing this support now, and dropping the dates module in this PR.

dao/src/main/java/se/fortnox/reactivewizard/db/ObservableStatementFactory.java
```diff
@@ -0,0 +1,114 @@
...
+        result = result.onBackpressureBuffer(RECORD_BUFFER_SIZE);
+        result = metrics.measure(result, time -> logSlowQuery(time, args));
+        result = LoggingContext.transfer(result);
+        result = result.subscribeOn(scheduler);
+
+        return new DaoObservable<>(result, transactionHolder);
+    }
+
+    private void logSlowQuery(long time, Object[] args) {
+        if (time > QUERY_LOG_THRESHOLD_MS) {
```
owtroid: This has been changed in the internal repo.

splitfeed: We'll do another PR to sync changes that occured during migration afterwards, it'll be easier :)

PR #58, opened 2018-05-04
Injection of ConfigFactory

rassmate:
Injection of ConfigFactory instead of explicit creation


config/src/main/java/se/fortnox/reactivewizard/config/ConfigFactory.java
```diff
@@ -2,11 +2,23 @@
 
 import com.fasterxml.jackson.databind.JsonNode;
 
+import javax.inject.Inject;
+import javax.inject.Named;
+
 public class ConfigFactory {
```
JonasHallFnx: @Singleton to avoid that the config is read one time for each injection

PR #60, opened 2018-05-07
Removed logging of slow query when in transaction

mickej:


splitfeed:
Why don't we want slow query logging in transactions? Isn't it just as relevant?

mickej:
The most relevant is SELECT queries, and those are not ran in a transaction. Sometimes we have pretty large transactions with thousands of INSERT/UPDATE, each INSERT takes just a few milliseconds. But the slow query logging takes the whole transaction when measuring the time, so it ends up with thousands of log statements.

JonasHallFnx:
@splitfeed the logging is per query, and a query is not considered completed until it is committed (it's Observable signals onCompleted). Therefore when we have a large transaction, all statements are marked as slow (they all complete at the same time) and we get a flood of logs.

Also, we often want to optimize reads more than writes. A transaction can never contain a read in our platform.

splitfeed:
Alright, so this is more of a workaround then? Just wanted to understand the change :)

PR #62, opened 2018-05-08
Enable config file passed in to TestInjector

rassmate:
Missed to allow the TestInjector to accept a config file as argument to fit my purpose

### DIFF
--- a/config/src/test/java/se/fortnox/reactivewizard/config/TestTestInjector.java
+++ b/config/src/test/java/se/fortnox/reactivewizard/config/TestTestInjector.java
@@ -11,7 +11,7 @@ public class TestTestInjector {
 
 
     @Test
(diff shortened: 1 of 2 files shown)


config/src/test/java/se/fortnox/reactivewizard/config/TestTestInjector.java
```diff
@@ -0,0 +1,34 @@
+package se.fortnox.reactivewizard.config;
+
+import com.google.inject.Injector;
+import com.google.inject.Key;
+import com.google.inject.name.Names;
+import org.junit.Test;
+
+import static org.fest.assertions.Assertions.assertThat;
+
+public class TestTestInjector {
```
splitfeed: What are the methods testing here? Improved method naming would be helpful :)

PR #63, opened 2018-05-08
Added validation module

JonasHallFnx:
With support for validating config files and resource calls

### DIFF
--- a/validation/src/main/java/se/fortnox/reactivewizard/validation/ValidatingConfigFactory.java
+++ b/validation/src/main/java/se/fortnox/reactivewizard/validation/ValidatingConfigFactory.java
@@ -1,0 +1,23 @@
+package se.fortnox.reactivewizard.validation;
+
+import se.fortnox.reactivewizard.config.ConfigFactory;
+
+import javax.inject.Inject;
+import javax.inject.Named;
+
+public class ValidatingConfigFactory extends ConfigFactory {
+    private final ValidatorUtil validatorUtil;
+
+    @Inject
+    public ValidatingConfigFactory(@Named("args") String[] args, ValidatorUtil validatorUtil) {
+        super(args);
+        this.validatorUtil = validatorUtil;
+    }
+
+    @Override
+    public <T> T get(Class<T> cls) {
+        T config = super.get(cls);
+        validatorUtil.validate(config);
+        return config;
+    }
+}
--- a/validation/src/main/java/se/fortnox/reactivewizard/validation/ValidatingConfigFactory.java
+++ b/validation/src/main/java/se/fortnox/reactivewizard/validation/ValidatingConfigFactory.java
@@ -4,7 +4,9 @@ import se.fortnox.reactivewizard.config.ConfigFactory;
 
 import javax.inject.Inject;
 import javax.inject.Named;
+import javax.inject.Singleton;
 
+@Singleton
 public class ValidatingConfigFactory extends ConfigFactory {
     private final ValidatorUtil validatorUtil;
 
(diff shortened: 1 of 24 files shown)


validation/src/main/java/se/fortnox/reactivewizard/validation/ValidatingProxy.java
```diff
@@ -0,0 +1,67 @@
...
+
+    @Override
+    public Object invoke(Object proxy, Method method, Object[] args) throws Throwable {
+        if (args != null) {
+            validatorUtil.validateParameters(target, method, args);
+        }
+        try {
+            return method.invoke(target, args);
+        } catch (ValidationFailedException e) {
+            throw e;
```
rassmate: Unnecessary catch since you just throw the exception.

validation/src/main/java/se/fortnox/reactivewizard/validation/ValidatingProxy.java
```diff
@@ -0,0 +1,67 @@
...
+        } catch (InvocationTargetException e) {
+            if (e.getTargetException() instanceof RuntimeException) {
+                throw (RuntimeException) e.getTargetException();
```
rassmate: Unnecessary cast

validation/src/main/java/se/fortnox/reactivewizard/validation/ValidatingProxy.java
```diff
@@ -0,0 +1,67 @@
...
+     * Wrapped implementation
+     * @return the wrapped instance
+     */
+    @Override
+    public Object get() {
+        return target;
+    }
+
+    /**
+     * Creates a proxy for the given service, whihc validates arguments before
```
rassmate: typo whihc = which

validation/src/main/java/se/fortnox/reactivewizard/validation/ValidatorUtil.java
```diff
@@ -0,0 +1,52 @@
...
+import java.lang.reflect.Method;
+import java.lang.reflect.Proxy;
+import java.util.Set;
+
+import javax.inject.Inject;
+import javax.validation.ConstraintViolation;
+import javax.validation.Validation;
+import javax.validation.Validator;
+import javax.validation.executable.ExecutableValidator;
+
```
rassmate: Since the validator used is singleton this validator util could be singleton as well

validation/src/main/java/se/fortnox/reactivewizard/validation/ValidationModule.java
```diff
@@ -0,0 +1,59 @@
...
+    private final InjectAnnotatedScanner injectAnnotatedScanner;
+
+    @Inject
+    public ValidationModule(InjectAnnotatedScanner injectAnnotatedScanner) {
+        this.injectAnnotatedScanner = injectAnnotatedScanner;
+    }
+
+    @Override
+    public void configure(Binder binder) {
+        binder.bind(Validator.class).toProvider(() -> {
```
rassmate: Maybe remove this binding and also have guice always use the default constructor in ValidatorUtil instead of the explicit @Inject constructor. Now we have the same validator configuration in two places.

validation/src/main/java/se/fortnox/reactivewizard/validation/ValidatingConfigFactory.java
```diff
@@ -0,0 +1,23 @@
+package se.fortnox.reactivewizard.validation;
+
+import se.fortnox.reactivewizard.config.ConfigFactory;
+
+import javax.inject.Inject;
+import javax.inject.Named;
+
```
rassmate: Should be singleton right, the singleton annotation is not inherited

validation/src/main/java/se/fortnox/reactivewizard/validation/ValidationModule.java
JonasHallFnx: Wouldn't that make it harder for others to override the configuration of the Validator? They would need to also change the binding of ValidatorUtil to use a provider which uses the other constructor?

rassmate: Agree, lets keep it this way.

PR #64, opened 2018-05-18
Feature/fast dao deserialize

JonasHallFnx:
This PR uses the power of LambdaMetafactory to replace reflection calls with compiled lambdas.

To encourage the separation of the bootstrapping and the execution of code using reflection, I have deprecated old methods like `getValue(Object object, String property)`. You should instead use `getter(Class<I> instanceCls, String propertyPath)` followed by a call to the returned Function.

I chose to use the existing interfaces in `java.util.function`, so that any reflection returned can be passed to code that does not need any reference to our utils module.


dao/src/test/java/se/fortnox/reactivewizard/db/deserializing/DeserializationPerformanceTest.java
```diff
@@ -0,0 +1,1391 @@
...
+            return false;
+        }
+    }
+
+    public static class MockedResultSet implements ResultSet {
+
+        private final int limit;
+        private int count;
+        private ResultSetMetaData metadata;
+        private String[] stringValues = new String[]{
```
splitfeed: Won't this test be optimized by the JVM when containing identical values for every entry, therefore being faster here than in real life? I'm thinking something like Faker might be useful.

JonasHallFnx: Yes, this test was removed in a later commit. I decided to avoid tests for the performance.

utils/src/main/java/se/fortnox/reactivewizard/util/FieldSetter.java
```diff
@@ -2,29 +2,30 @@
...
     private final Field    field;
     private final Class<?> parameterType;
     private final Type     genericParameterType;
 
-    private FieldSetter(Field field, Class<?> parameterType, Type genericParameterType) {
+    private FieldSetter(Field field, Class<T> parameterType, Type genericParameterType) {
         this.field = field;
         this.parameterType = parameterType;
         this.genericParameterType = genericParameterType;
         field.setAccessible(true);
```
rassmate: Should we for performance and to be consistent create a lambda function in the constructor here as well and use that lambda in the methods invoke and setterFunction. Otherwise we will return a new function each time setterFunction is used right?

JonasHallFnx: Good catch! Fixed.

utils/src/main/java/se/fortnox/reactivewizard/util/ReflectionUtil.java
```diff
@@ -259,6 +246,7 @@ private static Setter getSetter(Class<?> original, Class<?> declaringClass, Stri
                 return MethodSetter.create(original, first.get());
             }
         } catch (Exception ignored) {
+            ignored.printStackTrace();
```
owtroid: printStackTrace?

json/src/main/java/com/fasterxml/jackson/databind/ser/LambdaWriter.java
```diff
@@ -0,0 +1,164 @@
...
+        resolveLambda();
+        return this;
+    }
+
+
+    @Override
+    @SuppressWarnings("checkstyle:LocalVariableName")
+    public void serializeAsField(Object bean, JsonGenerator gen, SerializerProvider prov) throws Exception {
+        Object value = this.lambda != null ? this.lambda.apply(bean) :
+                this._accessorMethod == null ? this._field.get(bean) : this._accessorMethod.invoke(bean);
```
splitfeed: Wouldn't this nested ternary be more readable like this?

```java
        Object value;
        if (this.lambda != null) {
            value = this.lambda.apply(bean);
        } else if (this._accessorMethod == null) {
            value = this._field.get(bean);
        } else {
            value = this._accessorMethod.invoke(bean);
        }
```

json/src/main/java/com/fasterxml/jackson/databind/ser/LambdaWriter.java
```diff
@@ -0,0 +1,164 @@
...
+        }
+
+        // END COPY & PASTE
+        ////////////////////////////////
+
+    }
+
+    @Override
+    public void serializeAsElement(Object bean, JsonGenerator gen, SerializerProvider prov) throws Exception {
+        Object value = this.lambda != null ? this.lambda.apply(bean) :
```
splitfeed: Same as in serializeAsField regarding nested ternaries

utils/src/main/java/se/fortnox/reactivewizard/util/PropertyResolver.java
```diff
@@ -0,0 +1,230 @@
...
+            Property[]       newProperties         = new Property[this.properties.length + propertiesToAppend.length];
+            System.arraycopy(properties, 0, newProperties, 0, properties.length);
+            System.arraycopy(propertiesToAppend, 0, newProperties, properties.length, newProperties.length);
+
+            return Optional.of(new PropertyResolver(otherPropertyResolver.getPropertyType(), otherPropertyResolver.getPropertyGenericType(), newProperties));
+        }
+        return Optional.empty();
+    }
+
+    private void verifySetter(Property property) {
```
splitfeed: This method doesn't seem to be used anywhere, or is it?

utils/src/main/java/se/fortnox/reactivewizard/util/ReflectionUtil.java
```diff
@@ -349,4 +306,72 @@ private static Method getMethod(Class<?> cls, String propertyName) {
             return null;
         }
     }
+
+    public static <T> Optional<Supplier<T>> instantiator(Class<T> cls) {
```
splitfeed: Unless I'm mistaken, this does pretty much the same as newInstance did but with a more obscure name. Would it be wrong to call it newInstance or createInstance or something more descriptive like that?

JonasHallFnx: It does not, it returns a function that can be used to instantiate a class, as opposed to an instance, which was returned from newInstance.

splitfeed:
Nice updates! 👍

PR #66, opened 2018-06-04
Check for -parameters option and give a more helpful error message.

splitfeed:
Since this is now required we are seeing more and more issues with this missing and this is meant to help people realizing what the problem is before having to spend time debugging it.

The module checks that we can read a parameter name and if it fails it exists and prints an error message linking to our (first) wiki page: https://github.com/FortnoxAB/reactive-wizard/wiki/CompilerParameters

Please review the wiki page as well and update with anything that might be wrong or missing :)


dao/src/main/java/se/fortnox/reactivewizard/db/precheck/PrecheckModule.java
```diff
@@ -0,0 +1,38 @@
...
+    private static Logger LOG = LoggerFactory.getLogger(PrecheckModule.class);
+
+    public PrecheckModule() {
+        try {
+            Method    method         = PrecheckModule.class.getMethod("configure", Binder.class);
+            Parameter firstParameter = method.getParameters()[0];
+            if (!firstParameter.isNamePresent()) {
+                LOG.error("The project was not compiled with the -parameters option which is required for DAO parameter resolving. " +
+                    "Please see https://github.com/FortnoxAB/reactive-wizard/wiki/CompilerParameters for more information.");
+                System.exit(1);
```
splitfeed: Should this be done differently? I figured we would want to exit to make sure the service doesn't start and people miss the error, since it won't work properly anyway.

JonasHallFnx: Nice! Much needed! Any way this can be incorporated into the existing db module? Feels like it adds some costs to the already slow binding by acting as a separate no-op module.

splitfeed: Something like this? :)

Issue #70, opened 2018-06-11
HttpClientTest shouldCloseConnectionOnTimeoutDuringContentReceive is flaky


splitfeed:
The test randomly seems to fail both locally and on Travis, but I fail to see any obvious reason for it.

```
org.mockito.exceptions.verification.TooLittleActualInvocations: 
consumer.accept(
    "/hello/servertest/slowBody"
);
Wanted 11 times:
-> at se.fortnox.reactivewizard.client.HttpClientTest.shouldCloseConnectionOnTimeoutDuringContentReceive(HttpClientTest.java:562)
But was 10 times:
-> at se.fortnox.reactivewizard.client.HttpClientTest.lambda$createTestServer$12(HttpClientTest.java:509)
```

PR #81, opened 2018-08-29
Consider also "maven.surefire.debug" environment variable while decid…

skivol:
…ing if we are debugging

* to make debugging a more seamless process if started from command line using `maven`.

### DIFF
--- a/utils/src/main/java/se/fortnox/reactivewizard/util/DebugUtil.java
+++ b/utils/src/main/java/se/fortnox/reactivewizard/util/DebugUtil.java
@@ -4,17 +4,17 @@ package se.fortnox.reactivewizard.util;
  * Utilities for debugging in the development environment.
  */
 public class DebugUtil {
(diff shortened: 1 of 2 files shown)


splitfeed:
I don't know if or how you can verify the codecov checks without pushing, but the failing tests are due to some flakyness. I reported one of these in #70, but it doesn't seem to be limited to that single test any longer.

skivol:
Can we run the tests once again? :)

skivol:
Any ideas on how to make "codecov/patch" happy? (don't know if https://github.com/kata-containers/tests/issues/508 suggest a proper approach, but anyway :) )

splitfeed:
It looks strange, the coverage seems good to me. I'll just merge this :)

skivol:
Thanks :)

skivol:
@splitfeed, so when is the next release? :)

PR #82, opened 2018-08-31
Consider also "maven.surefire.debug" environment variable while decid…

skivol:
…ing if we are debugging

* to make debugging a more seamless process if started from command line using `maven`.
* trying to improve test coverage for "DebugUtil" by covering both branches for `isIdePresent` method.

(squashed previous 3 commits and want to check if the coverage/complexity checks pass)


skivol:
The build environment doesn't seem to feel good at the moment...

PR #83, opened 2018-08-31
Add custom entity and field validators

johannlindell:


johannlindell:
Discovered that another validation class is needed. Updating PR when everything is in place

JonasHallFnx:
@johannlindell I deliberately removed these classes when moving validation to github, because the validation module now knows about (has dependency on) most of the platform, rather than each module having dependencies towards validation. Therefore, putting this code in validation makes your code use most of the system, through inherited dependencies, which is not at all desireable.

I have no good solution to this problem, but it was a problem large enough for me to remove the class.

johannlindell:
@JonasHallFnx I see, then I'll look for another solution in my specific case

JonasHallFnx:
Closing due to inactivity

PR #85, opened 2018-09-05
urlEncode query parameters

MartinStengard:
Description:
We've experienced weird HttpClient errors when using HttpClientProvider to access external third party api's, in all environments. It's hard to recreate error in tests, succeeded a couple of times then error magically disappeared. 
URISyntaxException are thrown in encode(...) when having colon (:) in query parameters since query parameters are parsed through java.net.URI. Colon are not permitted in paths. In most cases it works fine but in deployed environments it's not working for us.

Error:
java.net.URISyntaxException: Illegal character in scheme name at index ...

Fix:
Always use existing urlEncode(...) for all annotation of QueryParam instance.


JonasHallFnx:
@sk8ter didn't you have a problem with slashes in query params in your project? Just having a small dejavu.

PR #89, opened 2018-09-25
Presence of a logging property will attempt to reconfigure JUL

magnusp:
A truthy julLoggingActivation will attempt to reconfigure Java utils logging from an configuration file, with a preset name, on the classpath.

### DIFF
--- a/logging/src/main/java/se/fortnox/reactivewizard/logging/LoggingFactory.java
+++ b/logging/src/main/java/se/fortnox/reactivewizard/logging/LoggingFactory.java
@@ -42,6 +42,9 @@ public class LoggingFactory {
     @JsonProperty("julLoggingActivation")
     Boolean julLoggingActivation = false;
 
+    /**
+     * Configure logging framework(s) and supply defaults.
+     */
     public void init() {
         if (appenders == null) {
             appenders = new HashMap<>();
@@ -61,17 +64,15 @@ public class LoggingFactory {
         }
         LoggingConfiguration.getInstance().configure(props);
 
-        /*
-        Provides a facility to reconfigure JUL via a reactivewizard config.
-         */
-        if(julLoggingActivation) {
-	        configureJUL();
+        // Provides a facility to reconfigure JUL via a reactivewizard config.
+        if (julLoggingActivation) {
+            configureJavaUtilLogging();
         }
     }
 
-    private void configureJUL() {
+    private void configureJavaUtilLogging() {
         final InputStream stream = LoggingFactory.class.getClassLoader().getResourceAsStream("rw-jul-logging.properties");
-        if(stream != null) {
+        if (stream != null) {
             try {
                 getLogManager().readConfiguration(stream);
             } catch (IOException e) {
--- a/logging/src/main/java/se/fortnox/reactivewizard/logging/LoggingFactory.java
+++ b/logging/src/main/java/se/fortnox/reactivewizard/logging/LoggingFactory.java
@@ -14,6 +14,8 @@ import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Properties;
 
+import static java.util.logging.LogManager.getLogManager;
+
 /**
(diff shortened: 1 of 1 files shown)


splitfeed:
What is the use case here, what does it do?

logging/src/main/java/se/fortnox/reactivewizard/logging/LoggingFactory.java
```diff
@@ -54,6 +60,22 @@ public void init() {
...
+        /*
+        Provides a facility to reconfigure JUL via a reactivewizard config.
+         */
+        if(julLoggingActivation) {
+	        configureJUL();
+        }
+    }
+
+    private void configureJUL() {
+        final InputStream stream = LoggingFactory.class.getClassLoader().getResourceAsStream("rw-jul-logging.properties");
```
splitfeed: Is rw-jul-logging.properties a file that should have been in the PR?

rassmate: And if the file is not present the startup will fail I guess. So perhaps include this getResourceAsStream row in the try catch statement

magnusp:
> What is the use case here, what does it do?

It allows users to apply a JUL configuration via a reactivewizard config. Internally this allows us to have one configuration file during development (standard JUL configuration) and one during production (using a formatter that fits Logstash/Elasticsearch). By having a hard coded filename users will only have to concern themselves with a choice between activating a configuration or not. Again, internally, we have such a file that will be put on the classpath.

JUL doesnt have facilities to load from the classpath, but rather will have to be provided with an absolute path to the file or a class name to configure logging. External users can thus choose to put either rw-jul-logging.properties on the classpath and activate it with reactivewizard, or run the jar with a path set on "java.util.logging.config.file" and let JUL configure everything.

logging/src/main/java/se/fortnox/reactivewizard/logging/LoggingFactory.java
magnusp: Regarding rw-jul-logging.properties, see a pull request hidden from the public eyes :D

logging/src/main/java/se/fortnox/reactivewizard/logging/LoggingFactory.java
```diff
@@ -36,6 +38,13 @@
     @JsonProperty("levels")
     Map<String, String> levels = new HashMap<>();
 
+    @Valid
```
JonasHallFnx: "Valid" annotations is used to validate sub-entities, like the items of a list. Does not apply to Boolean.

logging/src/main/java/se/fortnox/reactivewizard/logging/LoggingFactory.java
JonasHallFnx: ...but shouldn't there be a default config that does not override the formatter?

logging/src/main/java/se/fortnox/reactivewizard/logging/LoggingFactory.java
```diff
@@ -54,6 +63,22 @@ public void init() {
...
+        }
+    }
+
+    private void configureJavaUtilLogging() {
+        final InputStream stream = LoggingFactory.class.getClassLoader().getResourceAsStream("rw-jul-logging.properties");
+        if (stream != null) {
+            try {
+                getLogManager().readConfiguration(stream);
+            } catch (IOException e) {
+                e.printStackTrace();
```
JonasHallFnx: We probably want to throw this to avoid application from starting without correct logging?

logging/src/main/java/se/fortnox/reactivewizard/logging/LoggingFactory.java
magnusp: JUL has a default configuration under <path to JRE>/lib/logging.properties which I have no idea when it activates :)

PR #94, opened 2018-10-10
Check dao classes to be compiled with -parameters flag

rassmate:
Changed to System.err.println since it will be more visible to the developer when running mvn commands.


dao/src/main/java/se/fortnox/reactivewizard/db/DbProxyFactory.java
```diff
@@ -33,26 +30,34 @@ public DbProxyFactory(DaoClassScanner daoClassScanner) {
...
                 .in(Scopes.SINGLETON);
         });
     }
 
     /**
-     * Checks that we can access the name of a method parameter, and thereby
-     * verifying that the compilation whas run with -parameters.
+     * Checks if the -parameters flag was used to compile the dao class
+     * @param cls the class to use when checking for parameters
+     * @return null if no method with parameters was found, true if method found and -parameters flag was used
```
splitfeed: The method has void defined as return type, so this can't be right. Right?

PR #95, opened 2018-10-22
PLAT-1578 - Make readTimeout configurable

rassmate:
1. Making the readTimeout on nettys HttpClient configurable instead of always 10 seconds
2. Making the HttpClient using the configured timeout as the default observable timeout.

### DIFF
--- a/client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
+++ b/client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
@@ -69,8 +69,8 @@ public class HttpClient implements InvocationHandler {
     private final   Set<PreRequestHook>         preRequestHooks;
     private         RxClientProvider            clientProvider;
     private         ObjectMapper                objectMapper;
-    private         int                         timeout = 10000;
-    private         TimeUnit                    timeoutUnit = TimeUnit.MILLISECONDS;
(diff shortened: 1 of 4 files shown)


client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
```diff
@@ -70,8 +69,8 @@
     private final   Set<PreRequestHook>         preRequestHooks;
     private         RxClientProvider            clientProvider;
     private         ObjectMapper                objectMapper;
-    private         int                         timeout     = 10;
-    private         TimeUnit                    timeoutUnit = TimeUnit.SECONDS;
+    private         int                         timeout;
+    private         TimeUnit                    timeoutUnit = TimeUnit.MILLISECONDS;
```
magnusp: static final?

rassmate:
No can do. You can still change these settings by calling the setTimeout
method

Den tis 23 okt. 2018 kl 09:52 skrev Magnus Persson <notifications@github.com

client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
rassmate: Sorry can't do that.
We still want to be able to change these settings by calling the setTimeout method

PR #100, opened 2018-11-16
Register shutdown hook to process all connections before terminating …

JonasHallFnx:
…the system

### DIFF
--- a/server/src/main/java/se/fortnox/reactivewizard/server/ConnectionCounter.java
+++ b/server/src/main/java/se/fortnox/reactivewizard/server/ConnectionCounter.java
@@ -1,0 +1,40 @@
+package se.fortnox.reactivewizard.server;
+
+import javax.inject.Singleton;
+import java.util.concurrent.Semaphore;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicLong;
+
+@Singleton
+public class ConnectionCounter {
+    private final AtomicLong connections = new AtomicLong(0);
+    private final Semaphore connectionsZero = new Semaphore(1);
+
+    public void increase() {
+        if (connections.getAndIncrement() == 0) {
+            connectionsZero.tryAcquire();
+        }
+    }
+
+    public void decrease() {
+        if (connections.decrementAndGet() == 0) {
+            connectionsZero.release();
+        }
+    }
+
+    public boolean awaitZero(int time, TimeUnit timeUnit) {
+        try {
+            boolean success = connectionsZero.tryAcquire(time, timeUnit);
+            if (success) {
+                connectionsZero.release();
+            }
+            return success;
+        } catch (Exception e) {
+            return false;
+        }
+    }
+
+    public long getCount() {
+        return connections.get();
+    }
+}
--- a/server/src/main/java/se/fortnox/reactivewizard/server/RxNettyServer.java
+++ b/server/src/main/java/se/fortnox/reactivewizard/server/RxNettyServer.java
@@ -58,9 +58,9 @@ public class RxNettyServer extends Thread {
 
     private void registerShutdownHook() {
         Runtime.getRuntime().addShutdownHook(new Thread(() -> {
-            logger.info("Shutdown requested. Will wait up to 120 seconds...");
+            logger.info("Shutdown requested. Will wait up to {} seconds...", config.getShutdownTimeoutSeconds());
             server.shutdown();
-            if (!connectionCounter.awaitZero(120, TimeUnit.SECONDS)) {
+            if (!connectionCounter.awaitZero(config.getShutdownTimeoutSeconds(), TimeUnit.SECONDS)) {
                 logger.error("Shutdown proceeded while connection count was not zero: " + connectionCounter.getCount());
             }
             server.awaitShutdown();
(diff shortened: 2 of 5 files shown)


server/src/main/java/se/fortnox/reactivewizard/server/RxNettyServer.java
```diff
@@ -48,4 +55,17 @@ public void run() {
     public HttpServer<ByteBuf, ByteBuf> getServer() {
         return server;
     }
+
+    private void registerShutdownHook() {
+        Runtime.getRuntime().addShutdownHook(new Thread(() -> {
+            logger.info("Shutdown requested. Will wait up to 120 seconds...");
+            server.shutdown();
+            if (!connectionCounter.awaitZero(120, TimeUnit.SECONDS)) {
```
tsfcode: Borde inte timeouten vara configurerbar?
Körs servern exempelvis i kubernetes är det stor risk att servern blir dödad innan timeouten.

JonasHallFnx: Jo, det kan jag fixa. Default tid där är väl 30 sekunder va? Då kanske denna borde defaulta till 20 sekunder, för att undvika att få en kill -9?

tsfcode: Yes, det är 30 sekunder som är default för kubernetes. Så 20 sekunder låter rimligt :)

JonasHallFnx:
@splitfeed it is kind of hard to verify that, since it is a shutdown hook, which will be triggered when the JVM is shutting down.

server/src/main/java/se/fortnox/reactivewizard/server/ConnectionCounter.java
```diff
@@ -0,0 +1,40 @@
...
+    private final Semaphore connectionsZero = new Semaphore(1);
+
+    public void increase() {
+        if (connections.getAndIncrement() == 0) {
+            connectionsZero.tryAcquire();
+        }
+    }
+
+    public void decrease() {
+        if (connections.decrementAndGet() == 0) {
```
owtroid: Not a problem that will happen in this PR but what if someone were to call decrease() before an increase() has been called for some reason? Then the lock won't be acquired on the next call to increase() and awaitZero() won't actually wait since it will be able to acquire the lock directly.

JonasHallFnx: Not sure what scenario you are thinking of. Is it this?:
decrease()
increase()
increase()
decrease()

That one works just fine...

owtroid: No I was thinking:
decrease()  //this is done when connections = 0, by mistake or some weird reason
increase()
awaitZero()
.
.
decrease()

If the above happens awaitZero() will return immediately even though there is an outstanding request still going on.

magnusp:
> @splitfeed it is kind of hard to verify that, since it is a shutdown hook, which will be triggered when the JVM is shutting down.

So an application can be in different stages of its life and "shutting down" is one of them. Are we not really looking for the possibility of notifying observers when the application transitions to this state? And that a shutdown hook (or something else like an JMX kill command or as part of a test) is what initiates that transition?

server/src/main/java/se/fortnox/reactivewizard/server/ConnectionCounter.java
JonasHallFnx: So your scenario is that the first decrease() is faulty, rather than just being out of order? I guess we cannot support both those scenarios?

With current implementation we get:

**Scenario: first is faulty**
start, count = 0
decrease(), count = -1
increase(), count = 0
decrease(), count = -1
awaitZero() will fail

**Scenario: first is out of order**
start, count = 0
decrease(), count = -1
increase(), count = 0
increase(), count = 1
decrease(), count = 0
awaitZero() passes


If we change to support "faulty" calls:

**Scenario: first is faulty**
start, count = 0
decrease(), count = 0 (instead of -1)
increase(), count = 1
decrease(), count = 0
awaitZero() passes

**Scenario: first is out of order**
start, count = 0
decrease(), count = 0 (should have been -1)
increase(), count = 1
increase(), count = 2
decrease(), count = 1
awaitZero() will fail

So, as I see it, we cannot support both cases,and will have to choose the most likely of these unexpected scenarios. I would say it is more likely to have a race and ending up with calls out of order, than having spontaneus method calls (sunflares?).

JonasHallFnx:
> > @splitfeed it is kind of hard to verify that, since it is a shutdown hook, which will be triggered when the JVM is shutting down.
> 
> So an application can be in different stages of its life and "shutting down" is one of them. Are we not really looking for the possibility of notifying observers when the application transitions to this state? And that a shutdown hook (or something else like an JMX kill command or as part of a test) is what initiates that transition?

We could build an abstraction that supports adding multiple listeners that are called when the system is shutting down, but at this stage we don't need that extra complexity. It can be a nice refactor when we have multiple shutdown hooks.

Even if we would do that, it would be hard to _really_ verify that the hook will be called on shutdown, which is the problem here I think.

server/src/main/java/se/fortnox/reactivewizard/server/ConnectionCounter.java
owtroid: I was more thinking of a hypothetical case where a user of the class misbehaves. As mentioned, that won't happen with the changes in this PR and probably not ever. I don't think we need to worry about quantum effects spontaneously calling the method :)

splitfeed:
PR #100 btw! 🥇

PR #101, opened 2018-11-22
Feature/improved readme

JonasHallFnx:


config/src/main/java/se/fortnox/reactivewizard/config/ConfigFactory.java
```diff
@@ -12,15 +14,18 @@
...
 
     public ConfigFactory(String configFile) {
-        if (configFile != null && configFile.endsWith(".yml")) {
-            tree = ConfigReader.readTree(configFile);
-        } else {
+        if (configFile == null) {
+            tree = new POJONode(new Object());
+            return;
+        }
+        if (!configFile.endsWith(".yml")) {
```
splitfeed: This isn't really a new problem, but YAML commonly uses .yaml as file ending as well. In case someone happens to go by this code again, maybe we could add that :)

PR #102, opened 2018-11-26
Fixed so that retry will work together with transactional dao calls

jepp3:
I discovered that the DaoTransactions could not be retried. The implementation was throwing an "alreadyExecutedException" instead of doing a retry.  Now we will reset the state on error.

### DIFF
--- a/dao/src/main/java/se/fortnox/reactivewizard/db/transactions/Transaction.java
+++ b/dao/src/main/java/se/fortnox/reactivewizard/db/transactions/Transaction.java
@@ -4,6 +4,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import rx.Observable;
 import se.fortnox.reactivewizard.db.ConnectionProvider;
+import se.fortnox.reactivewizard.db.statement.Statement;
 
 import java.sql.Connection;
 import java.sql.SQLException;
@@ -57,8 +58,12 @@ public class Transaction<T> {
         } catch (Throwable e) {
             rollback(connection);
             closeConnection(connection);
-            allFailed(connection, e);
+            allFailed(e);
         }
+    }
+
+    public AtomicBoolean getWaitingForExecution() {
+        return waitingForExecution;
     }
 
     private void executeTransaction(Connection connection) throws SQLException {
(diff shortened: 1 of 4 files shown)


dao/src/main/java/se/fortnox/reactivewizard/db/transactions/Transaction.java
```diff
@@ -57,10 +58,14 @@ public void subscribed() {
         } catch (Throwable e) {
             rollback(connection);
             closeConnection(connection);
-            allFailed(connection, e);
+            allFailed(e);
         }
     }
 
+    public AtomicBoolean getWaitingForExecution() {
```
JonasHallFnx: not used?

dao/src/main/java/se/fortnox/reactivewizard/db/transactions/Transaction.java
```diff
@@ -94,16 +99,25 @@ private void closeConnection(Connection connection) {
...
                 log.error("onError threw exception", onErrorException);
             }
         }
     }
 
+    private void setTransactionToExecutable(TransactionStatement transactionStatement, Statement statement) {
+        final AtomicReference<Statement> atomicStatement = transactionStatement.getAtomicStatement();
+        final Transaction                transaction     = transactionStatement.getTransaction();
+        atomicStatement.compareAndSet(statement,null);
+        transaction.getWaitingForExecution().compareAndSet(false,true);
```
JonasHallFnx: Why compare and set if you are not using the result? Why doing this for each statement?

dao/src/main/java/se/fortnox/reactivewizard/db/transactions/Transaction.java
```diff
@@ -94,16 +99,25 @@ private void closeConnection(Connection connection) {
...
             } catch (Exception onErrorException) {
                 log.error("onError threw exception", onErrorException);
             }
         }
     }
 
+    private void setTransactionToExecutable(TransactionStatement transactionStatement, Statement statement) {
+        final AtomicReference<Statement> atomicStatement = transactionStatement.getAtomicStatement();
+        final Transaction                transaction     = transactionStatement.getTransaction();
+        atomicStatement.compareAndSet(statement,null);
```
JonasHallFnx: If statement.onError would do this reset, then the imlplementation details would not have to bubble here and this whole method would not be needed.

jepp3: I do not have the necessary information in the statement on error method ( or I might miss something here). But the method has been removed and the code left is easier to read.

PR #103, opened 2018-11-27
Prefer a secure HTTP client when calling https services

magnusp:
- Introduces an option "unsafeSecure" (default false) to make insecure requests.
- Allow the user to somewhat configure the SSL engine when there is
  a need to do so (custom trustmanagers, not having a keystore on
  file, etc).

### DIFF
--- a/client/src/main/java/se/fortnox/reactivewizard/client/HttpClientConfig.java
+++ b/client/src/main/java/se/fortnox/reactivewizard/client/HttpClientConfig.java
@@ -45,9 +45,11 @@ public class HttpClientConfig {
     private long maxRequestTime          = TimeUnit.MILLISECONDS.convert(1, MINUTES);
 
     private boolean isHttps;
-    private int     retryCount    = 3;
-    private int     retryDelayMs  = 1000;
-    private int     readTimeoutMs = 10000;
+    private int     retryCount     = 3;
+    private int     retryDelayMs   = 1000;
+    private int     readTimeoutMs  = 10000;
+    @JsonProperty("unsafeSecure")
+    private boolean isUnsafeSecure = false;
 
     public HttpClientConfig() {
     }
@@ -162,5 +164,13 @@ public class HttpClientConfig {
 
     public void setReadTimeoutMs(int readTimeoutMs) {
         this.readTimeoutMs = readTimeoutMs;
+    }
+
+    public boolean isUnsafeSecure() {
+        return isHttps() && isUnsafeSecure;
+    }
+
+    public void setUnsafeSecure(boolean value) {
+        isUnsafeSecure = value;
     }
 }
--- a/client/src/main/java/se/fortnox/reactivewizard/client/HttpClientConfig.java
+++ b/client/src/main/java/se/fortnox/reactivewizard/client/HttpClientConfig.java
@@ -45,11 +45,11 @@ public class HttpClientConfig {
     private long maxRequestTime          = TimeUnit.MILLISECONDS.convert(1, MINUTES);
 
     private boolean isHttps;
-    private int     retryCount     = 3;
-    private int     retryDelayMs   = 1000;
-    private int     readTimeoutMs  = 10000;
-    @JsonProperty("unsafeSecure")
-    private boolean isUnsafeSecure = false;
+    private int     retryCount    = 3;
+    private int     retryDelayMs  = 1000;
+    private int     readTimeoutMs = 10000;
+    @JsonProperty("insecure")
+    private boolean isInsecure    = false;
 
     public HttpClientConfig() {
     }
@@ -166,11 +166,11 @@ public class HttpClientConfig {
         this.readTimeoutMs = readTimeoutMs;
     }
 
-    public boolean isUnsafeSecure() {
-        return isHttps() && isUnsafeSecure;
+    public boolean isInsecure() {
+        return isHttps() && isInsecure;
     }
 
-    public void setUnsafeSecure(boolean value) {
-        isUnsafeSecure = value;
+    public void setInsecure(boolean value) {
+        isInsecure = value;
     }
 }
--- a/client/src/main/java/se/fortnox/reactivewizard/client/HttpClientConfig.java
+++ b/client/src/main/java/se/fortnox/reactivewizard/client/HttpClientConfig.java
@@ -45,11 +45,11 @@ public class HttpClientConfig {
     private long maxRequestTime          = TimeUnit.MILLISECONDS.convert(1, MINUTES);
 
     private boolean isHttps;
-    private int     retryCount    = 3;
-    private int     retryDelayMs  = 1000;
-    private int     readTimeoutMs = 10000;
-    @JsonProperty("insecure")
-    private boolean isInsecure    = false;
+    private int     retryCount             = 3;
+    private int     retryDelayMs           = 1000;
+    private int     readTimeoutMs          = 10000;
+    @JsonProperty("validateCertificates")
+    private boolean isValidateCertificates = true;
 
     public HttpClientConfig() {
     }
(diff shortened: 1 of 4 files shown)


client/src/test/java/se/fortnox/reactivewizard/client/HttpClientTest.java
```diff
@@ -938,6 +939,68 @@ public void shouldExecutePreRequestHooks() throws URISyntaxException {
...
+        });
+        RxClientProvider rxClientProvider = injector.getInstance(RxClientProvider.class);
+
+        try {
+            rxClientProvider
+                .clientFor(new InetSocketAddress(httpClientConfig.getHost(), httpClientConfig.getPort()))
+                .createGet("/")
+                .flatMap(HttpClientResponse::discardContent)
+                .toBlocking()
+                .singleOrDefault(null);
```
jepp3: add a fail in the end, so that test won't go green even if the exception isn't thrown.

client/src/main/java/se/fortnox/reactivewizard/client/RxClientProvider.java
```diff
@@ -54,11 +75,24 @@ public RxClientProvider(HttpClientConfig config, HealthRecorder healthRecorder)
...
+    /**
+     * Allows for customization of the SSLEngine
+     *
+     * @return An configured SSLEngine
+     */
+    protected SSLEngine configureSslEngine(String host, int port) {
+        try {
+            return SSLContext.getDefault().createSSLEngine(host, port);
+        } catch (Throwable e) {
+            throw new RuntimeException("Unable to configure secure client", e);
```
jepp3: maybe "https client" or something? So when it appears in the logs, it will be easier to understand what kind of client we are talking about. Just a maybe.

client/src/main/java/se/fortnox/reactivewizard/client/RxClientProvider.java
```diff
@@ -54,11 +75,24 @@ public RxClientProvider(HttpClientConfig config, HealthRecorder healthRecorder)
...
         }
         return client;
     }
 
+    /**
+     * Allows for customization of the SSLEngine
+     *
+     * @return An configured SSLEngine
+     */
+    protected SSLEngine configureSslEngine(String host, int port) {
```
jepp3: private or package protected

client/src/main/java/se/fortnox/reactivewizard/client/RxClientProvider.java
jepp3: add test case for the exception maybe. The current testcases will not run through this code as i can see.

client/src/main/java/se/fortnox/reactivewizard/client/HttpClientConfig.java
```diff
@@ -45,9 +45,11 @@
     private long maxRequestTime          = TimeUnit.MILLISECONDS.convert(1, MINUTES);
 
     private boolean isHttps;
-    private int     retryCount    = 3;
-    private int     retryDelayMs  = 1000;
-    private int     readTimeoutMs = 10000;
+    private int     retryCount     = 3;
+    private int     retryDelayMs   = 1000;
+    private int     readTimeoutMs  = 10000;
+    @JsonProperty("unsafeSecure")
```
splitfeed: I get the idea of naming the config like the setting in rxnetty, but isn't it a bit akward? I imagine something like "validateCertificates" or something like that would be easier to understand if you're not familiar with rxnetty.

client/src/main/java/se/fortnox/reactivewizard/client/HttpClientConfig.java
```diff
@@ -163,4 +165,12 @@ public int getReadTimeoutMs() {
     public void setReadTimeoutMs(int readTimeoutMs) {
         this.readTimeoutMs = readTimeoutMs;
     }
+
+    public boolean isUnsafeSecure() {
```
splitfeed: There's  a test for HttpClientConfig in HttpClientConfigTest, make sure to get coverage for this there as well :)

client/src/main/java/se/fortnox/reactivewizard/client/RxClientProvider.java
```diff
@@ -41,6 +44,24 @@ public RxClientProvider(HttpClientConfig config, HealthRecorder healthRecorder)
         return clients.computeIfAbsent(serverInfo, this::buildClient);
     }
 
+    private HttpClient<ByteBuf, ByteBuf> configureSsl(HttpClient<ByteBuf, ByteBuf> client, String host, int port, boolean isUnsafeSecure) {
+        if (isUnsafeSecure) {
+            return client.unsafeSecure();
+        }
+
+        try {
+            SSLEngine sslEngine = sslEngines.computeIfAbsent(host + ":" + port, s -> {
```
splitfeed: s?

client/src/main/java/se/fortnox/reactivewizard/client/HttpClientConfig.java
magnusp: unsafeSecure -> insecure, how does that sound? Taken from curl docs on the -k parameter

client/src/main/java/se/fortnox/reactivewizard/client/RxClientProvider.java
```diff
@@ -41,6 +44,24 @@ public RxClientProvider(HttpClientConfig config, HealthRecorder healthRecorder)
...
+                SSLEngine innerSslEngine = configureSslEngine(host, port);
+                innerSslEngine.setUseClientMode(true);
```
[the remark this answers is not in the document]
magnusp: Nope, this selects the expected initial handshake:
> Each SSL/TLS connection must have one client and one server, thus each endpoint must decide which role to assume. This choice determines who begins the handshaking process as well as which type of messages should be sent by each party. The method setUseClientMode(boolean) configures the mode. Once the initial handshaking has started, an SSLEngine can not switch between client and server modes, even when performing renegotiations.

client/src/main/java/se/fortnox/reactivewizard/client/HttpClientConfig.java
splitfeed: Like this? It doesn't feel obvious that it has to do with certificate validation, a lot of things about http clients could be insecure ;)
Also, insecure is a negation of secure which makes it even less clear.

httpClient:
  insecure: false?

client/src/main/java/se/fortnox/reactivewizard/client/RxClientProvider.java
magnusp: Private would not allow for customization so package protected is then preferred. Users would then need to put their variant of RxClientProvider under "our" namespace but I cannot motivate why that would be a bad thing - package protected it is!

client/src/main/java/se/fortnox/reactivewizard/client/RxClientProvider.java
magnusp: Im expanding the naming to hostPortValue, but it is in effect ignored. I couldn't find any better way while still using the nice computeIfAbsent.

client/src/main/java/se/fortnox/reactivewizard/client/RxClientProvider.java
magnusp: createSSLEngine is public final which I'm having trouble mocking to throw an exception. I believe exceptions would be thrown here if the JRE doesnt have a default keystore or if commandline parameters for the keystore/truststore point to non-existant files.

client/src/main/java/se/fortnox/reactivewizard/client/HttpClientConfig.java
magnusp: Ah, I see. I will change to "validateCertificates" and use a default of true.

PR #109, opened 2018-11-29
Add database socket timeout

erljef:
These changes update the version of HikariCP and add a DatabaseConfig option to configure a socket timeout.

Is the default timeout of 5 minutes reasonable or should we perhaps use 10 minutes? 
If any query takes longer to execute than this timeout the connection will be closed.


erljef:
Yes, I do believe the slow query warning will show even if an error occurs (such as the connection closing).

PR #110, opened 2018-11-30
Added factory method for SSL Engine.

jepp3:
Now a new SSL Engine will be created for each new connection. Otherwise the sslEnginge will be in a state where it thinks it has done a handshake when it in fact has not, since the connection is new.


client/src/main/java/se/fortnox/reactivewizard/client/RxClientProvider.java
```diff
@@ -48,15 +47,8 @@ public RxClientProvider(HttpClientConfig config, HealthRecorder healthRecorder)
...
-
         try {
-            SSLEngine sslEngine = sslEngines.computeIfAbsent(host + ":" + port, hostPortValue -> {
-                SSLEngine innerSslEngine = configureSslEngine(host, port);
-                innerSslEngine.setUseClientMode(true);
-                return innerSslEngine;
-            });
-
-            return client.secure(sslEngine);
+            return client.secure((ignored) -> createSslEngineForEachRequest(host, port));
```
JonasHallFnx: For each connection, I would say

PR #111, opened 2018-12-01
Added tests for server module

jepp3:
1) Added test cases for the server mode
2) Added excludes in pom file for classes that we don't need to test.
3) Moved LoggingMockUtil to test module.


server/src/main/java/se/fortnox/reactivewizard/server/ConnectionCounter.java
```diff
@@ -7,8 +7,18 @@
 
 @Singleton
 public class ConnectionCounter {
-    private final AtomicLong connections = new AtomicLong(0);
-    private final Semaphore connectionsZero = new Semaphore(1);
+    private final AtomicLong connections;
+    private final Semaphore connectionsZero;
+
+    public ConnectionCounter() {
+        connections = new AtomicLong(0);
```
JonasHallFnx: Perhaps use the protected constructor to avoid divergation later?

JonasHallFnx:
GJ!

server/src/main/java/se/fortnox/reactivewizard/server/ConnectionCounter.java
jepp3: sounds like a good idea!

PR #112, opened 2018-12-10
changed to warn logging

rassmate:
Changed to warn level.


rassmate:
Not necessary since 
https://github.com/FortnoxAB/reactive-wizard/pull/117
works around the real problem

PR #114, opened 2018-12-17
warn log if empty response

rassmate:
Logging information when a response is empty

### DIFF
--- a/client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
+++ b/client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
@@ -56,6 +56,8 @@ import java.util.concurrent.TimeoutException;
 import static java.lang.String.format;
 import static java.util.Collections.emptySet;
 import static javax.ws.rs.core.HttpHeaders.CONTENT_TYPE;
+import static rx.Observable.defer;
+import static rx.Observable.empty;
 import static rx.Observable.just;
 
 public class HttpClient implements InvocationHandler {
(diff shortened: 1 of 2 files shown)


client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
```diff
@@ -188,6 +196,15 @@ private void logFailedRequest(RequestBuilder fullReq, Throwable throwable) {
             .flatMap(str -> deserialize(method, str));
     }
 
+    private Observable<? extends ByteBuf> logEmptyResponse(RequestBuilder request, HttpClientResponse<ByteBuf> response) {
+        return defer(() -> {
+
```
jepp3: can we use the loggers built in supplant function instead?

splitfeed:
Wont this result in a lot of logs for endpoints that simply return no body, like 204:s or just empty 200:s?

JonasHallFnx:
> Wont this result in a lot of logs for endpoints that simply return no body, like 204:s or just empty 200:s?

I agree, we should check that expected returntype is not Observable<Void> I think.

JonasHallFnx:
Have you seen this? https://github.com/ReactiveX/RxNetty/issues/511

client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
rassmate: Since  I need to return an observable here I think not

PR #117, opened 2019-01-07
retry the request if submit returns empty (rare occations)

rassmate:
Workaround for occational emptys returned from the submit.
Will retry calls three times before actually returning empty.

### DIFF
--- a/client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
+++ b/client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
@@ -149,7 +149,7 @@ public class HttpClient implements InvocationHandler {
         final Observable<HttpClientResponse<ByteBuf>> resp = fullReq
             .submit(rxClient)
             .timeout(timeout, timeoutUnit)
-            .switchIfEmpty(logEmptyResponse(fullReq));
+            .switchIfEmpty(retry(fullReq, method, arguments, rxClient, true));
 
         Observable<?> output = null;
         if (expectsRawResponse(method)) {
@@ -167,6 +167,19 @@ public class HttpClient implements InvocationHandler {
         } else {
             return output;
         }
+    }
+
+    /**
+     * Creates a request builder based on the method and the arguments
+     *
+     * @param method    the method
+     * @param arguments the arguments
+     * @return an instance of RequestBuilder
+     */
+    private RequestBuilder getRequestBuilder(Method method, Object[] arguments) {
+        RequestBuilder fullReq = createRequest(method, arguments);
+        addDevOverrides(fullReq);
+        return fullReq;
     }
 
     private void logFailedRequest(RequestBuilder fullReq, Throwable throwable) {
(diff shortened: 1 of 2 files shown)


magnusp:
How does an resource returning Observable<Void> on GET behave? There will be no retries since headers are returned and thus we have a proper HttpClientResponse?

client/src/test/java/se/fortnox/reactivewizard/client/HttpClientTest.java
```diff
@@ -180,6 +223,76 @@ public void shouldRetryOnFullConnectionPool() {
...
+
+        try {
+
+            AtomicInteger callCount = new AtomicInteger();
+            TestResource resource = getHttpProxyWithClientReturningEmpty(callCount);
+            resource.getHello().toBlocking().singleOrDefault(null);
+            assertThat(callCount.get()).isEqualTo(3);
+
+        } catch (Exception e) {
+            Assert.fail("Should not give exception");
```
jepp3: Why catch? better to just add throws on the test case ?

client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
```diff
@@ -193,12 +206,48 @@ private void logFailedRequest(RequestBuilder fullReq, Throwable throwable) {
...
+        Object[] arguments,
+        io.reactivex.netty.protocol.http.client.HttpClient<ByteBuf, ByteBuf> rxClient,
+        boolean retryIfFails
+    ) {
         return defer(() -> {
 
-            LOG.warn(request + " with headers " + request.getHeaders().entrySet() + " got empty response from submit");
+            // Don't retry if it was a POST, as it is not idempotent and the request
+            // might have ended up at the server
+            if (request.getHttpMethod().equals(HttpMethod.POST)) {
```
jepp3: maybe getHttpMethod cannot be null, but feels safer the other way around.

client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
```diff
@@ -193,12 +206,48 @@ private void logFailedRequest(RequestBuilder fullReq, Throwable throwable) {
...
+
+            return fullReq
+                .submit(rxClient)
+                .timeout(timeout, timeoutUnit)
+                .switchIfEmpty(defer(() -> {
+
+                    if (retryIfFails) {
+                        LOG.warn("Got empty the second time as well. Giving it a third try");
+                        return retry(fullReq, method, arguments, rxClient, false);
+                    } else {
```
jepp3: maybe else is not needed?

client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
```diff
@@ -149,7 +149,7 @@ public Object invoke(Object proxy, Method method, Object[] arguments) {
         final Observable<HttpClientResponse<ByteBuf>> resp = fullReq
             .submit(rxClient)
             .timeout(timeout, timeoutUnit)
-            .switchIfEmpty(logEmptyResponse(fullReq));
+            .switchIfEmpty(retry(fullReq, method, arguments, rxClient, true));
```
JonasHallFnx: Is there a point in making the request from scratch or would it have worked with `.submit().single().retry(3)`?

client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
```diff
@@ -193,12 +206,48 @@ private void logFailedRequest(RequestBuilder fullReq, Throwable throwable) {
...
 
-            LOG.warn(request + " with headers " + request.getHeaders().entrySet() + " got empty response from submit");
+            // Don't retry if it was a POST, as it is not idempotent and the request
+            // might have ended up at the server
+            if (request.getHttpMethod().equals(HttpMethod.POST)) {
+                LOG.info("Won't retry POST on {} ", request.getUri());
+                return empty();
+            }
+
+            RequestBuilder fullReq = getRequestBuilder(method, arguments);
```
JonasHallFnx: Why create it again? You got it as a paramater to the method?

client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
rassmate: Doing this will bubble up the sequence contains no elements exception to the other framework retry handling exceptions and it will try config.retryCount() x config.retryCount() times.

client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
rassmate: Yes, thought at first it may hold some state and to avoid this problem i created a new instance. But will try reusing the old one.

client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
JonasHallFnx: ok, but Isn't that a good thing? That you have only one retry mechanism?

rassmate: I see where your aiming. Good point. Will try to fix tomorrow.

PR #131, opened 2019-02-21
WIP: Fix so that we are validating configurations

jepp3:
I found out that validation was not working on config object. The problem was that in Main.java in the bootstrap module, we define a configFactory that should be bound. We do the same in the autobind modules, but there we bind it to a configFactory with validation abilities. The AutoBindModules class says that the bootstrap bindings shall override the application bindings. The only way to get around the problem was to remove the binding from the bootstrap module.

### DIFF
--- a/bootstrap/src/main/java/se/fortnox/reactivewizard/Main.java
+++ b/bootstrap/src/main/java/se/fortnox/reactivewizard/Main.java
@@ -24,7 +24,6 @@ public class Main {
                 @Override
                 protected void configure() {
                     bind(String[].class).annotatedWith(Names.named("args")).toInstance(args);
-                    bind(ConfigFactory.class).toInstance(configFactory);
                 }
             };
             Guice.createInjector(new AutoBindModules(bootstrap));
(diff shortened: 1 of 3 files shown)


bootstrap/src/main/java/se/fortnox/reactivewizard/Main.java
```diff
@@ -24,7 +24,6 @@ public static void main(String[] args) {
                 @Override
                 protected void configure() {
                     bind(String[].class).annotatedWith(Names.named("args")).toInstance(args);
-                    bind(ConfigFactory.class).toInstance(configFactory);
```
JonasHallFnx: Without this, won't all AutoBindModules will get a ConfigFactory that is not instantiated correctly?

jepp3: I dont think so. I tested to set a breakpoint in the constructor and it loads the config file because the @Named("args") String[] args is bound. db queries seems to work properly indicating that the configuration is loaded properly and changing logging properties takes effect as well.

magnusp:
@jepp3 Any progress on this?

rassmate (approved):
Tried it out in one of our services and it failed correctly on startup when configuration was not valid. Ok to merge @jepp3

PR #135, opened 2019-03-04
Better error handling when streaming observables.

rassmate:
If there is an error in the observable that should be streamed then the headers are already written and can't be written again. 
In this pr I'll take the first element out of the stream and if that is ok then there is a good chance all the other elements will come as well. But if I cannot even get the first item I'll handle the error and writing the exception to the response instead.

### DIFF
--- a/jaxrs/src/main/java/se/fortnox/reactivewizard/jaxrs/response/JaxRsStreamingResult.java
+++ b/jaxrs/src/main/java/se/fortnox/reactivewizard/jaxrs/response/JaxRsStreamingResult.java
@@ -1,12 +1,19 @@
 package se.fortnox.reactivewizard.jaxrs.response;
 
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
 import io.netty.buffer.ByteBuf;
 import io.netty.handler.codec.http.HttpResponseStatus;
 import io.reactivex.netty.protocol.http.server.HttpServerResponse;
 import rx.Observable;
 import rx.functions.Func1;
+import se.fortnox.reactivewizard.jaxrs.WebException;
 
 import java.util.Map;
+
+import static rx.Observable.empty;
+import static rx.Observable.just;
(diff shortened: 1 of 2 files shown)


jaxrs/src/main/java/se/fortnox/reactivewizard/jaxrs/response/JaxRsStreamingResult.java
```diff
@@ -1,22 +1,55 @@
...
+                ObjectMapper objectMapper = new ObjectMapper();
+
+                headers.forEach(response::addHeader);
+
+                byte[] bytes;
+                try {
+                    bytes = objectMapper.writeValueAsBytes(throwable);
+                    return first(response.writeBytes(just(bytes))).thenReturn(empty());
+                } catch (JsonProcessingException e) {
+                    e.printStackTrace();
```
jepp3: remove print stack trace here. Log ?

jaxrs/src/main/java/se/fortnox/reactivewizard/jaxrs/response/JaxRsStreamingResult.java
```diff
@@ -1,22 +1,55 @@
...
         super(output, responseStatus, serializer, headers);
     }
 
     @Override
     public Observable<Void> write(HttpServerResponse<ByteBuf> response) {
-        response.setStatus(responseStatus);
-        headers.forEach(response::addHeader);
-        return response.writeBytesAndFlushOnEach(output.map(serializer));
+
+        return output
```
jepp3: if empty NoSuchElementException is thrown, maybe we should handle it without returning errror

jaxrs/src/main/java/se/fortnox/reactivewizard/jaxrs/response/JaxRsStreamingResult.java
```diff
@@ -1,22 +1,67 @@
...
+
+            .first()
```
JonasHallFnx: first kommer att unsubscriba efter att det kommit 1

PR #140, opened 2019-03-18
Support for serializing Lists as json

JonasHallFnx:
This allows for using a list as a value for a column.


dao/src/main/java/se/fortnox/reactivewizard/db/deserializing/DeserializerUtil.java
```diff
@@ -63,12 +62,7 @@
         BiFunction<PropertyResolver<I,?>, Deserializer, T> deserializerFactory
     ) {
         Type propertyGenericType = propertyResolver.getPropertyGenericType();
-        Function<String, ?> deserializer = JSON_DESERIALIZER_FACTORY.createDeserializer(new TypeReference<Object>() {
-            @Override
-            public Type getType() {
-                return propertyGenericType;
-            }
-        });
+        Function<String, ?> deserializer = JSON_DESERIALIZER_FACTORY.createDeserializer(propertyGenericType);
```
rassmate: Snyggt!

PR #141, opened 2019-03-20
Add host in headers in http client requests

johannlindell:
All request in HttpClient are sent with a host header. When a header param "Host" is specified, it is used,
otherwise the host from the http client config is used

### DIFF
--- a/client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
+++ b/client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
@@ -6,7 +6,6 @@ import com.fasterxml.jackson.databind.JsonMappingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.ObjectReader;
 import com.fasterxml.jackson.databind.type.TypeFactory;
-import com.netflix.servo.util.Strings;
 import io.netty.buffer.ByteBuf;
 import io.netty.handler.codec.http.HttpResponseStatus;
 import io.netty.handler.timeout.ReadTimeoutException;
--- a/client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
+++ b/client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
@@ -420,6 +420,10 @@ public class HttpClient implements InvocationHandler {
                 }
             }
         }
+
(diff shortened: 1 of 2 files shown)


client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
```diff
@@ -420,6 +420,10 @@ private void setHeaderParams(RequestBuilder request, Method method, Object[] arg
                 }
             }
         }
+
+        if (!request.getHeaders().containsKey("Host") || request.getHeaders().containsKey("Host") && request.getHeaders().get("Host").isEmpty()) {
```
JonasHallFnx: `isNullOrEmpty(request.getHeaders().get("Host"))`?

client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
```diff
@@ -6,6 +6,7 @@
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.ObjectReader;
 import com.fasterxml.jackson.databind.type.TypeFactory;
+import com.netflix.servo.util.Strings;
```
JonasHallFnx: I don't think we have an explicit dependency on servo... And we should eventually get rid of servo. If apache is not on the classpath then perhaps just write the code yourself?

client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
```diff
@@ -420,6 +421,10 @@ private void setHeaderParams(RequestBuilder request, Method method, Object[] arg
                 }
             }
         }
+        
+       if (isNullOrEmpty(request.getHeaders().get("Host"))) {
```
splitfeed: The build is failing because indentation is wrong here and on line 427, just add one more space and you're good to go 👍

client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
JonasHallFnx: The import is still here?

johannlindell: Sorry! Will remove

PR #142, opened 2019-03-28
Parse field errors from WebExceptions when handling errors in HttpClient

owtroid:
In case of a rw service calling another rw service that generates an error, the fielderrors from that error is lost when handling the error in HttpClient. 

This tries to make sure that the fielderrors are set on the DetailedError cause.

### DIFF
--- a/client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
+++ b/client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
@@ -15,8 +15,10 @@ import org.slf4j.LoggerFactory;
 import rx.Observable;
 import rx.Single;
 import se.fortnox.reactivewizard.jaxrs.ByteBufCollector;
+import se.fortnox.reactivewizard.jaxrs.FieldError;
 import se.fortnox.reactivewizard.jaxrs.JaxRsMeta;
 import se.fortnox.reactivewizard.jaxrs.WebException;
+import se.fortnox.reactivewizard.json.JsonDeserializerFactory;
 import se.fortnox.reactivewizard.logging.LoggingContext;
 import se.fortnox.reactivewizard.metrics.HealthRecorder;
 import se.fortnox.reactivewizard.metrics.Metrics;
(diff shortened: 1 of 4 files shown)


client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
```diff
@@ -15,8 +15,10 @@
 import rx.Observable;
 import rx.Single;
 import se.fortnox.reactivewizard.jaxrs.ByteBufCollector;
+import se.fortnox.reactivewizard.jaxrs.FieldError;
 import se.fortnox.reactivewizard.jaxrs.JaxRsMeta;
 import se.fortnox.reactivewizard.jaxrs.WebException;
+import se.fortnox.reactivewizard.json.JsonDeserializerFactory;
```
JonasHallFnx: not used?

owtroid: Fixed

jaxrs-api/src/main/java/se/fortnox/reactivewizard/jaxrs/FieldError.java
```diff
@@ -3,14 +3,18 @@
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonInclude.Include;
 
+import java.io.Serializable;
 import java.util.Map;
 
-public class FieldError {
+public class FieldError implements Serializable {
```
JonasHallFnx: What benefit does this give?

owtroid: SonarCloud did not like that I added FieldError to a serializable class (DetailedError) without making it serializable or transient. Making it transient didn't feel right so I tried this instead but that just pushed the error into this class and the errorParams object instead :(

owtroid: SonarCloud also got mad that DetailedError is not immutable (it wants Throwables to be immutable) but it wasn't immutable before my changes either. What gives?

PR #145, opened 2019-05-17
Now logs the error on every retry in HttpClient

Spetastian:
In order to know what error was returned from an API we should log every error on every retry.

### DIFF
--- a/client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
+++ b/client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
@@ -66,8 +66,8 @@ import static rx.Observable.error;
 import static rx.Observable.just;
 
 public class HttpClient implements InvocationHandler {
-    private static final Logger           LOG            = LoggerFactory.getLogger(HttpClient.class);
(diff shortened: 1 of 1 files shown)


client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
```diff
@@ -254,6 +254,10 @@ private void addDevOverrides(RequestBuilder fullRequest) {
                     return false;
                 }
                 if (((WebException)throwable).getStatus().code() >= 500) {
+
+                    // Log the error on every retry.
+                    LOG.info("Will retry because an error occurred", throwable);
```
JonasHallFnx: If I'm not mistaken, this will not log anything about the request, you probably want to log something like `String request = format("%s, headers: %s", fullReq.getFullUrl(), fullReq.getHeaders().entrySet());`, or is this already part of the caught WebException?

Spetastian: It is my understanding that the caught WebException includes the response from the service and that is what we want to log.

PR #148, opened 2019-05-25
Support for @BeanParam when calling other services

JonasHallFnx:
This adds support for the @BeanParam when making http calls to external services using the HttpClient.

Since the HttpClient is not yet rewritten to make all reflection at startup, like the JaxRsResource-classes, I resorted to a simple cache to avoid doing the heavy reflection parts on each request.

Also added the Tuple class, feel free to use it wherever you like :-)


utils/src/main/java/se/fortnox/reactivewizard/util/Tuple.java
```diff
@@ -0,0 +1,25 @@
+package se.fortnox.reactivewizard.util;
+
+/**
+ * Generic helper for dealing with Tuples
+ * @param <A> what ever you need!
+ * @param <B> what ever you need!
+ */
+public class Tuple<A,B> {
+
+    private final A value1;
```
jepp3: Because both value1 and value 2 is final and the class is immutable, maybe its redundant to have getters at all? Maybe just have them public?

One could then write 
t.value1 
instead of
t.getValue1()

But maybe we have some static code check that will complain ( even if it would be safe )

client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
```diff
@@ -509,6 +519,16 @@ protected String getPath(Method method, Object[] arguments, JaxRsMeta meta) {
                     } else {
                         path = path.replaceAll("\\{" + ((PathParam)annotation).value() + "\\}", this.urlEncode(this.serialize(value)));
                     }
+                } else if (annotation instanceof BeanParam) {
```
jepp3: It might be a bit complex to mutate the list we are iterating over?

Would it be possible to do the "conversion of bean params" to normal params earlier, to make the code easier to understand? So that we break down the process into steps, instead of doing everything at once?

client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
```diff
@@ -509,6 +519,16 @@ protected String getPath(Method method, Object[] arguments, JaxRsMeta meta) {
...
+                    if (value == null) {
+                        continue;
+                    }
+                    beanParamCache
+                        .computeIfAbsent(types[i], this::getBeanParamGetters)
+                        .forEach(v -> {
+                            args.add(v.getValue1().apply(value));
```
jepp3: My first question when reading this line is " what type is Value1 and what type is value2 ?" Can we add them to local variables, just to make it easier to follow the code? Now when im reading the code, i need to go into the getBeanParamGetter method to see what it might return.

client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
```diff
@@ -518,12 +538,26 @@ protected String getPath(Method method, Object[] arguments, JaxRsMeta meta) {
...
+                result.add(new Tuple<>(
+                    getter.get(),
+                    field.getAnnotations()
+                ));
+            }
+        }
+        return result;
+    }
+
     protected String serialize(Object value) {
```
jepp3: is there a reason why we have protected methods? would it be safe to make them "package private" and not expose them outside of the package. Because i have a feeling that it might be what we intended? But might be risky to risky to do?

utils/src/main/java/se/fortnox/reactivewizard/util/Tuple.java
JonasHallFnx: I think that the Tuple would be more useful with setters, so that it can mutate, but I didn't need that right now. I would rather add setters than go for using the fields.

client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
JonasHallFnx: Doing the conversion of BeanParams first would require to iterate the arguments twice. From my point of view, this implementation is simple, it iterates each argument and deals with each case one by one.

I plan to rewrite HttpClient in such a way that when the client is created, all reflection is done and at runtime we have a set of functions that will deal with the known arguments without a single if-statement. But that is not something I wanted to do in the same PR as adding this functionality.

client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
JonasHallFnx: That is indeed the downside of the Tuple... I will add local vars.

client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
JonasHallFnx: Having them protected would be to allow subclasses to change behaviour. It is easy to make this new method protected in the future, but going the other way is not possible, so I think private is right atm.

client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
```diff
@@ -80,6 +88,7 @@
     private         ObjectMapper                objectMapper;
     private         int                         timeout = 10;
     private         TimeUnit                    timeoutUnit = TimeUnit.SECONDS;
+    private final Map<Class<?>, List<Tuple<Function<Object, Object>, Annotation[]>>> beanParamCache = new HashMap<>();
```
splitfeed: What is the reason for adding a Tuple instead of having a proper class with names that explain what the values represent?

client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
```diff
@@ -509,6 +519,19 @@ protected String getPath(Method method, Object[] arguments, JaxRsMeta meta) {
...
+                } else if (annotation instanceof BeanParam) {
+                    if (value == null) {
+                        continue;
+                    }
+                    beanParamCache
+                        .computeIfAbsent(types[i], this::getBeanParamGetters)
+                        .forEach(v -> {
+                            Function<Object, Object> beanPropertyGetter = v.getValue1();
+                            Object beanPropertyValue = beanPropertyGetter.apply(value);
+                            Annotation[] annotations = v.getValue2();
```
splitfeed: Again regarding the Tuple, it makes this code very hard to follow. It also requires the developer to keep track of which value is saved in which of the values.

client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
```diff
@@ -509,6 +519,19 @@ protected String getPath(Method method, Object[] arguments, JaxRsMeta meta) {
...
+                            args.add(beanPropertyValue);
+                            argumentAnnotations.add(annotations);
```
splitfeed: This changing of the list we are looping, both args and argumentAnnotations makes this REALLY hard to follow

JonasHallFnx: Why is it hard to follow? I think it is quite simple. While processing one argument, some new arguments are added.

As I have said before, I plan to rewrite the logic in HttpClient to make it functional and to avoid all reflection at runtime, but I would like to do that in a separate PR.

client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
JonasHallFnx: Because adding a proper class is painful? I guess you're not a fan of the Tuple concept...

client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
JonasHallFnx: Disregarding this piece of code, should we add the Tuple class to utils or not? I had the need to simply return two values without any fuzz, and I've had this problem many times before, so this time I added the Tuple, thinking it would be useful in more places. Anyone else who has opinions about the Tuple concept?

client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
```diff
@@ -487,10 +496,11 @@ protected String getPath(Method method, Object[] arguments, JaxRsMeta meta) {
 
         StringBuilder  query       = null;
         Class<?>[]     types       = method.getParameterTypes();
-        Annotation[][] annotations = method.getParameterAnnotations();
-        for (int i = 0; i < types.length; i++) {
-            Object value = arguments[i];
-            for (Annotation annotation : annotations[i]) {
+        List<Object> args = new ArrayList<>(asList(arguments));
```
rassmate: args = new ArrayList<>(asList(arguments));
maybe just 
args = asList(arguments);

jepp3: Will not work if we do :
"args.add(v.getValue1().apply(value));"

asList returns immutable list

JonasHallFnx: Yes, would have been nice, but then I can't add new vales to the list, which happens when a @BeanParam is discovered, as that param will map one argument to multiple.

client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
rassmate: Aha...

PR #149, opened 2019-05-27
File server

JonasHallFnx:
Add this module to be able to serve files. Nothing fancy, intended for local development.


file-server/src/test/java/se/fortnox/reactivewizard/fileserver/FileServerTest.java
```diff
@@ -0,0 +1,35 @@
+package se.fortnox.reactivewizard.fileserver;
+
+import io.reactivex.netty.protocol.http.server.MockHttpServerRequest;
+import org.junit.Test;
+import rx.Observable;
+import se.fortnox.reactivewizard.mocks.MockHttpServerResponse;
+
+import static org.fest.assertions.Assertions.assertThat;
+
+public class FileServerTest {
```
jepp3: Maybe add some more unit test for the requesthandler when the path is not set in the config, as well when the index file is specified in the config and that utf-8 is specified when js files. Just to not reduce code coverage in project in overall. 

as well, the test class should have the same name as the class it tests

file-server/pom.xml
```diff
@@ -0,0 +1,40 @@
...
+            <version>${project.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>se.fortnox.reactivewizard</groupId>
+            <artifactId>reactivewizard-jaxrs</artifactId>
+            <version>${project.version}</version>
+            <scope>test</scope>
+            <classifier>tests</classifier>
+        </dependency>
```
jepp3: think coverage stuff should be specified per module, just to make sure it does not go down in the future.

JonasHallFnx: will add coverage checks

file-server/src/main/java/se/fortnox/reactivewizard/fileserver/FileServerRequestHandler.java
```diff
@@ -0,0 +1,121 @@
...
+        response.setStatus(HttpResponseStatus.OK);
+        for (Map.Entry<String, String> e : config.getHeaders().entrySet()) {
+            response.addHeader(e.getKey(), e.getValue());
+        }
+
+        setContentType(response, file);
+
+        long requestStartTime = System.currentTimeMillis();
+        Observable<byte[]> fileStream = FileUtil.readFile(file.toString(), 4096);
+        return merge(request.discardContent(), response.writeBytesAndFlushOnEach(fileStream)
```
magnusp: Is merge the correct operator here? Isn't there a risk that discardContent happens before FileUtil.readFile completes?

file-server/src/test/java/se/fortnox/reactivewizard/fileserver/FileServerTest.java
```diff
@@ -0,0 +1,35 @@
...
+
+    @Test
+    public void readsFile() {
+        FileServerConfig config = new FileServerConfig();
+        config.setPath(".");
+        FileServerRequestHandler fileServer = new FileServerRequestHandler(config);
+        MockHttpServerResponse response = new MockHttpServerResponse();
+        Observable<Void> result = fileServer.handle(new MockHttpServerRequest("/pom.xml"), response);
```
magnusp: Perhaps add a test that where the file is larger than the buffer of FileUtil.read? Or have the buffer configurable

file-server/src/main/java/se/fortnox/reactivewizard/fileserver/FileServerRequestHandler.java
rassmate: but the two observables works on different objects right? The first discards content from request and the latter writes the answer to the response. So I'd imagine it works as expected.

JonasHallFnx: correct, request and response have nothing to do with each other, but botch should be carried out.

PR #150, opened 2019-06-04
add the ability to point a resource to a certain httpClientConfig.

rassmate:
Before this if you needed a custom httpClient you needed the config, a new providerclass and a new binding in a module.
This way we don't need that and just need

@UseHttpClientConfig("configname")

on the resource.
This requires a subclass of HttpClientConfig present with @Config annotation with the same value


JonasHallFnx:
An important aspect of reactive wizard is the separation between resource interface and implementation, allowing us to write code that is the same no matter if the resource is local or remote. The API does not give any hint about how it is implemented.

Would it be possible to invert your idea, so that you instead put an annotation on the config, pointing to the resource interface? Or some other way of getting the benefits of your code while keeping the separation between interface and implementation?

client/src/main/java/se/fortnox/reactivewizard/client/RestClientFactory.java
```diff
@@ -4,41 +4,73 @@
...
+    private <T> Provider<T> provider(Class<T> iface,
+        Provider<HttpClientProvider> httpClientProvider, Provider<? extends HttpClientConfig> httpClientConfigProvider) {
         return () -> {
-            T httpProxy = httpClientProvider.get().create(iface);
-            log.debug("Created {} for {}", Proxy.getInvocationHandler(httpProxy), iface.getName());
+
+            HttpClientConfig httpClientConfig = httpClientConfigProvider.get();
+
+            //Create client based on config and create proxy
+            T httpProxy = httpClientProvider.get().createClient(httpClientConfig).create(iface);
```
JonasHallFnx: wouldn't this result in creating a new client (with a new pool and so on) for each resource?

client/src/main/java/se/fortnox/reactivewizard/client/RestClientFactory.java
```diff
@@ -4,41 +4,73 @@
...
+
+        Map<Class, Class<?>> httpClientConfigByResource = new HashMap<>();
+
+        httpConfigClassScanner
+            .getClasses()
+            .forEach(configClass -> {
+                if (configClass.isAnnotationPresent(UseInResource.class)) {
+                    for (Class resource : configClass.getAnnotation(UseInResource.class).value()) {
+                        if (!resource.isInterface()) {
+                            LOG.warn("{} pointed out in UseInResource must be an interface", resource);
```
JonasHallFnx: I don't usually watch the logs so close. Perhaps we should just throw an exception so the system won't start under this circumstance? Or can it ever be a valid case?

client/src/main/java/se/fortnox/reactivewizard/client/RestClientFactory.java
JonasHallFnx: ...perhaps HttpClientProvider can keep a cache with client per config?

JonasHallFnx:
I think it would be interesting to start a normal system and count the number of hits in the HttpClient constructor, to ensure it is equal to the number of unique configs, and not to the number of resources.

client/src/main/java/se/fortnox/reactivewizard/client/RestClientFactory.java
splitfeed: A fatal error sounds good to me :)
Perhaps the name of config could be included in the error message as well for easier debugging?

client/src/main/java/se/fortnox/reactivewizard/client/RestClientFactory.java
rassmate: Correct, I thought the old implementation did that as well since HttpClient was not annotated with Singleton but there the RxClientProvider is injected and that class is a Singleton.

So to mimic the old behaviour I will in HttpClientProvider create a cache with RxClientProviders using the config as a key

client/src/main/java/se/fortnox/reactivewizard/client/HttpClientProvider.java
```diff
@@ -28,7 +31,12 @@ public HttpClientProvider(HealthRecorder healthRecorder,
         this.preRequestHooks = preRequestHooks;
     }
 
-    public HttpClient createClient(HttpClientConfig config) {
-        return new HttpClient(config, new RxClientProvider(config, healthRecorder), objectMapper, requestParameterSerializers, preRequestHooks);
+    public HttpClient createClient(HttpClientConfig httpClientConfig) {
+
+        //Fill up cache with RxClientProviders
+        rxClientProviderCache.computeIfAbsent(httpClientConfig, config -> new RxClientProvider(config, healthRecorder));
```
JonasHallFnx: This will return the `RxClientProvider`, no need for an additional get below.

client/src/main/java/se/fortnox/reactivewizard/client/HttpClientConfig.java
```diff
@@ -190,4 +191,21 @@ public void setBasicAuth(String username, String password) {
             .setPassword(password);
     }
 
+    @Override
+    public boolean equals(Object object) {
```
JonasHallFnx: this might become a problem for classes extending this class. Why did you need this? Without this, you would create a new RxClientFactory for each config instance, but isn't there just a single instance anyway?

rassmate: My mistake.

rassmate: No, the configs will be a new instance for each resource.

JonasHallFnx: Perhaps we can change that by making this binding into a singleton? https://github.com/FortnoxAB/reactive-wizard/blob/master/config/src/main/java/se/fortnox/reactivewizard/config/ConfigAutoBindModule.java#L26

rassmate: Yes, that would have worked but now I went for a Class<? extends HttpClientConfig> as key instead.

PR #151, opened 2019-06-07
Change from fest-assertions to AssertJ

splitfeed:
Reactive Wizard has been using fest-assert that has not been updated since 2011, but there is a well maintained branch of it called AssertJ that is still activly being developed. Changes between them are minor and can usually be interchanged by just switching import.

The old dependency on fest-assert is left in place to not cause breaking changes, but allow people to start using AssertJ instead.

**This change is meant to be fully backwards compatible, so please let me know if something looks breaking.**


pom.xml
```diff
@@ -81,8 +81,8 @@
         <guava.version>27.0.1-jre</guava.version>
 
         <junit.version>4.12</junit.version>
-        <fest.version>1.4</fest.version>
         <fest-assert.version>1.4</fest-assert.version>
+        <assertj.version>3.11.1</assertj.version>
```
flowertwig: 3.12.2 seems to be the latest

config/src/test/java/se/fortnox/reactivewizard/config/ConfigReaderTest.java
```diff
@@ -158,7 +156,7 @@ public void shouldThrowExceptionForInvalidYaml() {
             ConfigReader.fromFile("src/test/resources/testconfig-invalid.yml", EmptyConfig.class);
             fail("Expected exception, but none was thrown");
         } catch (RuntimeException exception) {
-            assertException(exception, MarkedYAMLException.class);
+            assertThat(exception).hasRootCauseInstanceOf(MarkedYAMLException.class);
```
flowertwig: assertj allows us to get rid of try/catch, but that could be a later improvement:
`assertThatThrownBy(() -> ConfigReader.fromFile("src/test/resources/testconfig-invalid.yml", EmptyConfig.class)).isInstanceOf(RuntimeException.class).hasRootCauseInstanceOf(MarkedYAMLException.class);`

splitfeed: Yeah, I tried to avoid rewriting too much tests unless required to keep down scope. Good catch though :)

PR #155, opened 2019-06-28
Validation exceptions should not be loglevel WARN but INFO

rassmate:
Changing ValidationExceptions and WebExceptions with only FieldErrors to logLevel INFO


JonasHallFnx:
@rassmate can you merge this if it's ready?

PR #159, opened 2019-08-23
Do not serialize as json when a custom content type has been used

JonasHallFnx:


client/src/test/java/se/fortnox/reactivewizard/client/HttpClientTest.java
```diff
@@ -1292,6 +1292,80 @@ public void assertRequestContainsHostFromHeaderParam() {
...
+            return request.getContent().flatMap(buf -> {
+                recordedRequestBody.set(buf.toString(Charset.defaultCharset()));
+                return Observable.empty();
+            });
+        });
+
+        TestResource resource = getHttpProxy(server.getServerPort());
+
+        try {
+            resource.sendXml(new Pojo()).toBlocking().lastOrDefault(null);
```
flowertwig: fail if no exception

JonasHallFnx: fixed

flowertwig: AssertJ is nice for this btw :)
```
assertThatExceptionOfType(type)
.isThrownBy((call)
.withMessage(mess);
```

PR #162, opened 2019-09-18
Remove LoggingContext

flowertwig:
Remove LoggingContext


JonasHallFnx:
@flowertwig would it be possible to just remove the usage of the LoggingContext in this PR, and then we can perhaps have the new context stuff in our private RW until it has matured? This would give us faster turnaround times at this moment.

PR #169, opened 2019-11-22
Case insensitive matching when using IN/NOT IN

magnusp:


dao/src/test/java/se/fortnox/reactivewizard/db/CollectionOptionsTest.java
```diff
@@ -153,7 +153,7 @@ public void shouldAddOrderByBeforeQueryOrderByWithoutDefaultSort() throws Except
     public void shouldInjectSortBeforeQueriesLastOrderBy() throws SQLException {
         CollectionOptions collectionOptions = new CollectionOptions("name", CollectionOptions.SortOrder.ASC);
         collectionOptionsDao.selectWithMultipleOrderBy(collectionOptions).toBlocking().singleOrDefault(null);
-        mockDb.verifySelect("select name, lastname from customers where name in (select first_name from persons order by first_name) order by name ASC, lastname LIMIT 101");
+        mockDb.verifySelect("select name, lastname from customers where name = ANY(select first_name from persons order by first_name) order by name ASC, lastname LIMIT 101");
```
magnusp: @JonasHallFnx should be an ok change right?

PR #171, opened 2019-11-26
Update Mockito from 1.10.16 to 3.1.0

linsnos:
... and depending bytebuddy from 1.9.3 to 1.10.3.

### DIFF
--- a/client/src/test/java/se/fortnox/reactivewizard/client/RestClientFactoryTest.java
+++ b/client/src/test/java/se/fortnox/reactivewizard/client/RestClientFactoryTest.java
@@ -46,7 +46,7 @@ public class RestClientFactoryTest {
             mockResourcesOnClassPath(binder, of(TestResource.class));
 
             when(httpClientProvider.createClient(any(HttpClientConfig.class))).thenAnswer(invocation -> {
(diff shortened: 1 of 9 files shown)


flowertwig:
The mockito api has changed, so this will force changes. Just so we're aware.

flowertwig:
Also, some matching behaviour has changed, i.e. anyX and any(class).

JonasHallFnx:
> Also, some matching behaviour has changed, i.e. anyX and any(class).

@flowertwig Do you propose we upgrade to a lower version? The upgrade was needed to support mocking of internal Auth class having interface with default implementation, which is supported in version 2+.

flowertwig:
> > Also, some matching behaviour has changed, i.e. anyX and any(class).
> 
> @flowertwig Do you propose we upgrade to a lower version? The upgrade was needed to support mocking of internal Auth class having interface with default implementation, which is supported in version 2+.

No. The upgrade is long overdue and I'm pretty sure the biggest difference is between 1 and 2 anyway. We just need to be aware that there will be compilation failures and probably some failing tests.

rassmate:
Mockito has been upgraded in another pr. Closing this one

PR #173, opened 2019-11-29
cache metadata for outgoing requests

rassmate:
Since the JaxRsMeta is always the same for a particular method it seems wise to cache this and avoid a lot of processing and annotation parsing in each and every outgoing request.

### DIFF
--- a/client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
+++ b/client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
@@ -87,6 +87,7 @@ public class HttpClient implements InvocationHandler {
     private         int                         timeout = 10;
     private         TimeUnit                    timeoutUnit = TimeUnit.SECONDS;
     private final Map<Class<?>, List<BeanParamProperty>> beanParamCache = new HashMap<>();
+    private final Map<Method, JaxRsMeta>        jaxRsMetaMap = new HashMap<>();
 
     @Inject
     public HttpClient(HttpClientConfig config,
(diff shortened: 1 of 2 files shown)


client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
```diff
@@ -413,8 +414,14 @@ private DetailedError getDetailedError(String data, Throwable cause) {
         return detailedError;
     }
 
+    protected JaxRsMeta getJaxRsMeta(Method method) {
+        jaxRsMetaMap.computeIfAbsent(method, JaxRsMeta::new);
```
flowertwig: You can just return the result of compute...

client/src/test/java/se/fortnox/reactivewizard/client/HttpClientTest.java
```diff
@@ -274,6 +273,27 @@ public void shouldNotSometimesDropItems() {
...
+                return jaxRsMeta;
+            }
+        };
+
+        Method getHello = TestResource.class.getMethod("getHello");
+
+        httpClient.createRequest(getHello,new Object[0]);
+        httpClient.createRequest(getHello,new Object[0]);
+
+        assertThat(jaxRsMetas.get(0)).isSameAs(jaxRsMetas.get(1));
```
flowertwig: A test with another method would be nice, just to show it doesn't cache just one object. :)

client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
```diff
@@ -87,6 +87,7 @@
     private         int                         timeout = 10;
     private         TimeUnit                    timeoutUnit = TimeUnit.SECONDS;
     private final Map<Class<?>, List<BeanParamProperty>> beanParamCache = new HashMap<>();
+    private final Map<Method, JaxRsMeta>        jaxRsMetaMap = new HashMap<>();
```
JonasHallFnx: ConcurrentHashMap to avoid concurrency problems?

PR #174, opened 2019-11-29
Expose path

rassmate:
For a particular request we want to expose the JaxRsMeta path


client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
```diff
@@ -413,8 +414,14 @@ private DetailedError getDetailedError(String data, Throwable cause) {
         return detailedError;
     }
 
+    protected JaxRsMeta getJaxRsMeta(Method method) {
+        jaxRsMetaMap.computeIfAbsent(method, JaxRsMeta::new);
+        return jaxRsMetaMap.get(method);
```
JonasHallFnx: No need to have a get here, you can just return the row above

rassmate:
Got the branches mixed up. Creating another PR for the path exposure

PR #175, opened 2019-12-02
expose path

rassmate:
There is information about the full path when making a request.
But when answering a request the full path is not to be found.
This pr solves that so that you can get the path of the resource that will answer the request


jaxrs/src/main/java/se/fortnox/reactivewizard/jaxrs/JaxRsRequestHandler.java
```diff
@@ -11,6 +11,9 @@
 
 import javax.inject.Inject;
 import javax.inject.Singleton;
+import java.util.Optional;
```
flowertwig: not needed :P

PR #176, opened 2019-12-12
Shutdown Netty server gracefully.

sk8ter:
Shutdown Netty server gracefully.
Delay Netty shutdown before resolving shutdown dependency (if there is one).

> Shutting down a Netty application is usually as simple as shutting down all EventLoopGroups you created via shutdownGracefully(). It returns a Future that notifies you when the EventLoopGroup has been terminated completely and all Channels that belong to the group have been closed.

### DIFF
--- a/server/src/main/java/se/fortnox/reactivewizard/server/RxNettyServer.java
+++ b/server/src/main/java/se/fortnox/reactivewizard/server/RxNettyServer.java
@@ -1,6 +1,9 @@
 package se.fortnox.reactivewizard.server;
 
 import io.netty.buffer.ByteBuf;
+import io.netty.channel.EventLoopGroup;
+import io.netty.channel.socket.nio.NioServerSocketChannel;
+import io.reactivex.netty.RxNetty;
 import io.reactivex.netty.protocol.http.server.HttpServer;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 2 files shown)


server/src/main/java/se/fortnox/reactivewizard/server/RxNettyServer.java
```diff
@@ -65,17 +75,50 @@ public void run() {
     }
 
     private void registerShutdownHook() {
-        Runtime.getRuntime().addShutdownHook(new Thread(() -> shutdownHook(config, server, connectionCounter)));
+        Runtime.getRuntime().addShutdownHook(new Thread(() -> shutdownHook(config, server, eventLoopGroup, connectionCounter)));
+    }
+
+    public static void registerShutdownDependency(Runnable blockShutdownUntil) {
+        if (RxNettyServer.blockShutdownUntil != null && blockShutdownUntil != null) {
+            throw new IllegalStateException("Shutdown dependency is already registered");
```
JonasHallFnx: Perhaps you can have a list of Runnables so that we can have multiple shutdown hooks?

server/src/main/java/se/fortnox/reactivewizard/server/RxNettyServer.java
```diff
@@ -65,17 +75,50 @@ public void run() {
...
+        RxNettyServer.blockShutdownUntil = blockShutdownUntil;
     }
 
-    static void shutdownHook(ServerConfig config, HttpServer server, ConnectionCounter connectionCounter) {
-        LOG.info("Shutdown requested. Will wait up to {} seconds...", config.getShutdownTimeoutSeconds());
-        server.shutdown();
+    static void shutdownHook(ServerConfig config, HttpServer server, EventLoopGroup eventLoopGroup, ConnectionCounter connectionCounter) {
+        LOG.info("Shutdown requested.");
+        awaitShutdownDependency();
+        LOG.info("Will wait up to {} seconds...", config.getShutdownTimeoutSeconds());
```
JonasHallFnx: The log should probably be before the above row, and you should only allow waiting for the configured time.

server/src/main/java/se/fortnox/reactivewizard/server/RxNettyServer.java
```diff
@@ -65,17 +75,50 @@ public void run() {
...
+    static void awaitShutdownDependency() {
+        if (blockShutdownUntil == null) {
+            return;
+        }
+
+        LOG.info("Wait for completion of shutdown dependency");
+        Thread thread = new Thread(blockShutdownUntil);
+        thread.start();
+        try {
+            thread.join();
```
JonasHallFnx: This join needs to have a timelimit that is the time left until config.getShutdownTimeoutSeconds() has been reached.

server/src/main/java/se/fortnox/reactivewizard/server/RxNettyServer.java
```diff
@@ -65,17 +75,50 @@ public void run() {
...
+        }
+        RxNettyServer.blockShutdownUntil = blockShutdownUntil;
     }
 
-    static void shutdownHook(ServerConfig config, HttpServer server, ConnectionCounter connectionCounter) {
-        LOG.info("Shutdown requested. Will wait up to {} seconds...", config.getShutdownTimeoutSeconds());
-        server.shutdown();
+    static void shutdownHook(ServerConfig config, HttpServer server, EventLoopGroup eventLoopGroup, ConnectionCounter connectionCounter) {
+        LOG.info("Shutdown requested.");
+        awaitShutdownDependency();
```
JonasHallFnx: You should probably shutdown incoming requests before this is called, oatherwise the registered shutdown hook might get new work all the time?

server/src/main/java/se/fortnox/reactivewizard/server/RxNettyServer.java
JonasHallFnx: postponed.

server/src/main/java/se/fortnox/reactivewizard/server/RxNettyServer.java
JonasHallFnx: postponed, not easy/possible to shutdown only incoming requests

PR #179, opened 2020-01-31
Adding QueryParam annotations to the CollectionOptions class

rassmate:
so that classes used with @BeanParam annotations in resources can extend CollectionOptions to avoid extra parameters in resources

### DIFF
--- a/api/src/main/java/se/fortnox/reactivewizard/CollectionOptions.java
+++ b/api/src/main/java/se/fortnox/reactivewizard/CollectionOptions.java
@@ -14,7 +14,7 @@ public class CollectionOptions {
     @QueryParam("offset")
     private Integer   offset;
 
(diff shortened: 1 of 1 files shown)


api/src/main/java/se/fortnox/reactivewizard/CollectionOptions.java
```diff
@@ -1,14 +1,23 @@
...
         DESC
     }
 
+    @QueryParam("limit")
     private Integer   limit;
+
+    @QueryParam("offset")
     private Integer   offset;
+
+    @QueryParam("sortBy")
```
JonasHallFnx: Are these case senesitive? If so, this is not compatible with `CollectionOptionsResolver` we have internally which uses `sortby`.

PR #183, opened 2020-03-06
Add a request scoped context

erljef:
The context is open during handling of a JaxRsRequest, see changes to JaxRsRequestHandler. It is basically a Map stored in a ThreadLocal that is propagated through Observables using https://github.com/reactiverse/reactive-contexts.

This is a part in automatically propagating auth info through the system.


jaxrs/src/main/java/se/fortnox/reactivewizard/jaxrs/JaxRsRequestHandler.java
```diff
@@ -87,11 +92,16 @@ public JaxRsRequestHandler(Object[] services,
...
-        return resource.call(jaxRsRequest)
-            .singleOrDefault(null)
-            .flatMap(result -> writeResult(response, result))
-            .onErrorResumeNext(e -> exceptionHandler.handleException(request, response, e))
-            .doAfterTerminate(() -> resource.log(request, response, requestStartTime));
+        try {
+            JaxRsRequestContext.open();
+            return resource.call(jaxRsRequest)
+                .singleOrDefault(null)
+                .flatMap(result -> writeResult(response, result))
```
magnusp: If you add a test where writeResult leaves the current thread I think you would find that JaxRsRequestContext.close() would execute before the flatMap terminates.

erljef: The perils of RxJava. It is true that in the current thread (the thread that creates the observable) the context is closed before the observable executes. The reactive-contexts library solves that by propagating the context to the created observable (and any observables it in turn creates).

jaxrs/pom.xml
```diff
@@ -123,6 +123,18 @@
             <version>${project.version}</version>
             <scope>test</scope>
         </dependency>
+
+        <dependency>
+            <groupId>io.reactiverse</groupId>
+            <artifactId>reactive-contexts-propagators-rxjava1</artifactId>
+            <version>${reactiversecontexts.version}</version>
+            <exclusions>
+                <exclusion>
```
JonasHallFnx: Will there be a risk of not getting the version of rxjava needed to support contexts?

erljef: Good point. I don't believe it will (since the used api is available in the version we use, 1.2.10). However, since we're using 1.3.8 in our internal modules perhaps we should update the version here as well?

jaxrs/src/main/java/se/fortnox/reactivewizard/jaxrs/JaxRsRequestHandler.java
```diff
@@ -87,11 +92,16 @@ public JaxRsRequestHandler(Object[] services,
...
+                .onErrorResumeNext(e -> exceptionHandler.handleException(request, response, e))
+                .doAfterTerminate(() -> resource.log(request, response, requestStartTime));
+        } finally {
+            JaxRsRequestContext.close();
```
flowertwig: Move to doOnTerminate or it will be closed right after the observable is returned.
Some test for this would be nice.

jaxrs/pom.xml
```diff
@@ -123,6 +123,18 @@
             <version>${project.version}</version>
             <scope>test</scope>
         </dependency>
+
+        <dependency>
+            <groupId>io.reactiverse</groupId>
+            <artifactId>reactive-contexts-propagators-rxjava1</artifactId>
+            <version>${reactiversecontexts.version}</version>
+            <exclusions>
```
flowertwig: Is this because it uses a later version? Perhaps we should consider upping ours instead.

jaxrs-api/src/main/java/se/fortnox/reactivewizard/jaxrs/context/JaxRsRequestContext.java
```diff
@@ -0,0 +1,92 @@
...
+     * Open a new context.
+     *
+     * The context does not support reentry and will throw an exception if a context already exists.
+     * User code should not need to open a context.
+     *
+     * @throws IllegalStateException if a context already exists
+     */
+    public static void open() {
+        if (CONTEXT.get() != null) {
+            throw new IllegalStateException("A context already exists");
```
flowertwig: We should probably fail silently instead and log this. It would also be nice to have open and close as package private.

jaxrs-api/src/main/java/se/fortnox/reactivewizard/jaxrs/context/JaxRsRequestContext.java
```diff
@@ -0,0 +1,92 @@
...
+    /**
+     * Retrieve a value stored in the context.
+     *
+     * @param key The key associated with the value
+     * @param <T> The value type
+     * @return An Optional containing the value. The Optional is empty if there is no value set,
+     *         if the value set is null or if the context is not open.
+     */
+    public static <T> Optional<T> getValue(Object key) {
+        JaxRsRequestContext ctx = getContext();
```
flowertwig: Consider using map
`return ofNullable(getContext())
.map(context -> (T) context.contextStorage.get(key));`

flowertwig:
We also need this
`Don't forget to list your context propagator's fully-qualified class names in the META-INF/services/io.reactiverse.reactivecontexts.core.ContextPropagator file, and to include it in your classpath.`

jaxrs/src/main/java/se/fortnox/reactivewizard/jaxrs/JaxRsRequestHandler.java
erljef: That is by design. I will add a test to show how it works.

jaxrs-api/src/main/java/se/fortnox/reactivewizard/jaxrs/context/JaxRsRequestContext.java
erljef: I'm not sure. Since reentry is not supported and any failure would mean a programmatic error don't we want to fail spectacularly ASAP?

I thought about the visibility that but couldn't decide. It was easier having them public to make testing  easier. Will figure out a solution.

JonasHallFnx:
In our internal implementation of context it is totally pluggable, so you can add it as a module if you want it or omit it if you don't need it. It adds some memory overhead, and for certain usecases this feature is not needed.

erljef:
> In our internal implementation of context it is totally pluggable, so you can add it as a module if you want it or omit it if you don't need it. It adds some memory overhead, and for certain usecases this feature is not needed.

With internal context, do you mean the tracing context?
This request context would be part of the base functionality for keeping for example authorization info available throughout the request. A similar context will be available for event-/message handling to transfer information. The tracing module could be modified to use these contexts instead I think.

The features using the context are pluggable but I'm not sure if making the context feature itself pluggable is necessary.
Should it be pluggable by extending the JaxRsRequestHandler class (like the tracing module)? That would restrict the usage by only allowing either the tracing module or the context module (unless you combine them in user code somehow).

It could be made pluggable by implementing a pluggable wrapper around the call to JaxRsResource.call() in the handler (think ServletFilter) and have a jaxrs-context module implement that feature.

jaxrs-api/src/main/java/se/fortnox/reactivewizard/jaxrs/context/JaxRsRequestContext.java
flowertwig: So you decided to keep open/close public? I won't argue against, but just want to check that it wasn't forgotten.
Also, what's the reason for throwing exception when opening incorrectly, but not when closing?

erljef: Yeah, I think it should probably be visible.

Opening twice may cause unexpected problems that are not immediately visible since it will clear the current context much like if it was closed (meaning, the properties are gone). Closing an already closed context does nothing and thus doesn't throw an exception.

erljef:
After discussion with Jonas we agreed that the context functionality should be pluggable in itself.
The jaxrs module will expose an interface that other modules can implement to plug in extra handling around the call to JaxRsResource.call() and the context will be moved to a separate module and make use of the new plugin mechanism.

jaxrs/src/main/java/se/fortnox/reactivewizard/jaxrs/JaxRsResourceInterceptor.java
```diff
@@ -0,0 +1,62 @@
...
+/**
+ * Interface that allows for customized JaxRsResource call handling.
+ * Applications may register interceptors to add pre- or post processing behaviour common to all resource calls without
+ * having to modify the resources.
+ *
+ * The preHandle and postHandle methods will be run in the same thread but the execution of the call will run
+ * asynchronously. Any processing of the actual call should be made on the resourceCall observable in postHandle.
+ */
+public interface JaxRsResourceInterceptor {
+
```
rassmate: Seeing a future where we want to have a priority between interceptors but we could implement that when the need arises

erljef: My thoughts exactly =D

PR #185, opened 2020-04-02
Fix version

rassmate:


api/pom.xml
```diff
@@ -6,7 +6,7 @@
 	<parent>
 		<artifactId>reactivewizard-parent</artifactId>
 		<groupId>se.fortnox.reactivewizard</groupId>
-		<version>1.0.0-SNAPSHOT</version>
+		<version>9.9.9-SNAPSHOT</version>
```
JonasHallFnx, line 9: 999.0.0 perhaps so that this version will not be lower than the future release version 10.2.1?

binding/pom.xml
```diff
@@ -6,7 +6,7 @@
 	<parent>
 		<groupId>se.fortnox.reactivewizard</groupId>
 		<artifactId>reactivewizard-parent</artifactId>
-		<version>1.0.0-SNAPSHOT</version>
+		<version>9.9.9-SNAPSHOT</version>
```
JonasHallFnx, line 9: Perhaps use ${project.version} in all sub-poms?

PR #188, opened 2020-04-20
Stop retries for POST calls

johannlindell:
Make http method check first in withRetry() to stop all retries for POSTS

### DIFF
--- a/client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
+++ b/client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
@@ -56,7 +56,6 @@ import java.util.Date;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.NoSuchElementException;
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
@@ -265,6 +264,10 @@ public class HttpClient implements InvocationHandler {
     protected <T> Observable<T> withRetry(RequestBuilder fullReq, Observable<T> response) {
         return response.retryWhen(new RetryWithDelay(config.getRetryCount(), config.getRetryDelayMs(),
             throwable -> {
+                if (fullReq.getHttpMethod().equals(POST)) {
+                    // Don't retry if it was a POST, as it is not idempotent
+                    return false;
+                }
                 if (throwable instanceof TimeoutException) {
                     return false;
                 }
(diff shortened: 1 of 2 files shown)


client/src/test/java/se/fortnox/reactivewizard/client/HttpClientTest.java
```diff
@@ -139,6 +139,28 @@ public void shouldNotRetryFailedPostCalls() {
         }
     }
 
+    @Test
+    public void shouldNotRetryFailedPostCallsWithSystemExceptions() {
+        AtomicLong callCount = new AtomicLong();
+        HttpServer<ByteBuf, ByteBuf> server    = startServer(INTERNAL_SERVER_ERROR, "\"NOT OK\"", r -> callCount.incrementAndGet());
```
JonasHallFnx, line 145: perhaps add a delay and return 200 OK, to trigger the read timeout which is the error we are trying to fix?

client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
```diff
@@ -274,18 +277,9 @@ private String createBasicAuthString() {
                     return false;
                 }
                 if (!(throwable instanceof WebException)) {
-                    // Don't retry posts when a NoSuchElementException is raised since the request might have ended up at the server the first time
-                    if (throwable instanceof NoSuchElementException && POST.equals(fullReq.getHttpMethod())) {
-                        return false;
-                    }
-
                     // Retry on system error of some kind, like IO
                     return true;
```
JonasHallFnx, line 281: This was the row that triggered the retry before. It was designed this way, without early exit for POST, in order to retry POSTs if the reason is a local error, like a DNS lookup failure or something that means that the request has not left this system. I guess this is now more up to the caller, to add approriate retries for POSTs.

johannlindell: Do you think its better to "blacklist" exceptions for post call retries, so we just add the ReadTimeOutException in the condtion?

JonasHallFnx: No, I think you did the right thing here. Since we have had problems with requests being retried, rather than not being retried, I think this is the way to go. Blacklisting exceptions is hard here, there could be other low level exceptions that are hard to know if they ment the request was sent or not. So my comment was more of a clarification.

client/src/test/java/se/fortnox/reactivewizard/client/HttpClientTest.java
```diff
@@ -139,6 +139,35 @@ public void shouldNotRetryFailedPostCalls() {
         }
     }
 
+    @Test
+    public void shouldNotRetryFailedPostCallsWithSystemExceptions() {
+        AtomicLong callCount = new AtomicLong();
+        HttpServer<ByteBuf, ByteBuf> server = startServer(OK, "\"TAKES LONGER THAN READ TIMEOUT\"", r -> {
+            callCount.incrementAndGet();
+            try {
+                Thread.sleep(6);
```
JonasHallFnx, line 148: hehe, that's blocking code, dude! Search for other tests in this class "Slow server" to find examples of a proper slow server.

PR #189, opened 2020-04-20
Change transactional statements to avoid subscription on other thread

JonasHallFnx:
Before this change, each statement of a transaction was subscribed on the scheduler associated with the DbProxy. This is now changed so that the only execution on the scheduler will be the last statement in the transaction (the actual execution of the transaction).


JonasHallFnx:
I did affect some public methods. Do you think I should:
A) change the PR to keep it backwardscompatible
B) release this as a new major version
C) pretend that it is in fact an "internal" api and no one should have used those methods

JonasHallFnx:
> Missing some tests though to verify this.

@rassmate behaviour has not changed at all, all existing tests governing behaviour are still passing. The only thing that has changed is the number of threads allocated, which is hard to measure in a test because depending on the circumstances it may or may not mean that the existing io-scheduler threads are reused.

rassmate:
> > Missing some tests though to verify this.
> 
> @rassmate behaviour has not changed at all, all existing tests governing behaviour are still passing. The only thing that has changed is the number of threads allocated, which is hard to measure in a test because depending on the circumstances it may or may not mean that the existing io-scheduler threads are reused.

Fair enough!

PR #190, opened 2020-04-24
Use log4j to log startup errors

magnusp:
Effectively replace e.printStackTrace() with a logger call.

### DIFF
--- a/bootstrap/src/main/java/se/fortnox/reactivewizard/Main.java
+++ b/bootstrap/src/main/java/se/fortnox/reactivewizard/Main.java
@@ -4,9 +4,12 @@ import com.google.inject.AbstractModule;
 import com.google.inject.Guice;
 import com.google.inject.Module;
 import com.google.inject.name.Names;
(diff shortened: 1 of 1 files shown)


bootstrap/src/main/java/se/fortnox/reactivewizard/Main.java
```diff
@@ -29,8 +37,35 @@ protected void configure() {
...
+     *
+     * @param args commandline arguments
+     * @return A ConfigFactory based on a configuration file.
+     * @throws NoSuchFileException If the configuration file cannot be found.
+     */
+    private static ConfigFactory createConfigFactory(String[] args) throws NoSuchFileException {
+        try {
+            return new ConfigFactory(args);
+        } catch (RuntimeException runtimeException) {
+            if (runtimeException.getCause().getClass().isAssignableFrom(NoSuchFileException.class)) {
```
flowertwig, line 63: Nullcheck for cause?

PR #193, opened 2020-05-14
Release newly created scheduler workers after use

erljef:
When using Scheduler.io(), creating a worker for each transaction in
ObservableStatementFactory without releasing them never lets the
scheduler reuse workers and will create threads until there are no
more resources left.

### DIFF
--- a/dao/src/main/java/se/fortnox/reactivewizard/db/ObservableStatementFactory.java
+++ b/dao/src/main/java/se/fortnox/reactivewizard/db/ObservableStatementFactory.java
@@ -4,6 +4,8 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import rx.Observable;
 import rx.Scheduler;
+import rx.Subscriber;
+import rx.functions.Action0;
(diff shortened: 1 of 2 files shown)


dao/src/main/java/se/fortnox/reactivewizard/db/ObservableStatementFactory.java
```diff
@@ -111,6 +97,21 @@ private static void closeSilently(Connection connection) {
...
+        Scheduler.Worker worker = scheduler.createWorker();
+        worker.schedule(() -> {
+            try {
+                action.call();
+            } catch (Exception e) {
+                if (!subscription.isUnsubscribed()) {
+                    subscription.onError(e);
+                }
+            } finally {
+                worker.unsubscribe();
```
JonasHallFnx, line 110: This is the important part?

erljef: Yes, according to the documentation on Scheduler.createWorker(): "When work is completed it should be unsubscribed using Scheduler.Worker#unsubscribe()."

PR #195, opened 2020-06-15
Feature/reactor server

JonasHallFnx:
Replace RxNetty with ReactorNetty
    
This is a rather simple change, most changes are related to changing request/response classes from the ones RxNetty had to the ones Reactor has, but they pretty much work the same.
    
While doing this change I also scrapped the support for blocking return types. I don't see much use of that. If someone want's to do blocking stuff they should be aware of what they are doing and create a reasobale scheduler for that themselves.
    
In this commit the old client was replaced by what was in reactor-client.
    
The plan is to follow this up with a PR allowing you to use Flux, Mono, but also RxJava2 types perhaps.


utils/src/test/java/se/fortnox/reactivewizard/util/FluxTest.java
```diff
@@ -0,0 +1,27 @@
+package se.fortnox.reactivewizard.util;
```
rassmate, line 1: This class was just a local test i guess that got committed by accident

utils/src/main/java/se/fortnox/reactivewizard/util/ReflectionUtil.java
```diff
@@ -45,7 +47,7 @@ public static Type getTypeOfObservable(Method method) {
         }
         ParameterizedType parameterizedType = (ParameterizedType)type;
         Class<?>          rawClass          = (Class<?>)parameterizedType.getRawType();
-        if (!(rawClass.equals(Observable.class)) && !(rawClass.equals(Single.class))) {
+        if (!(rawClass.equals(Observable.class)) && !(rawClass.equals(Single.class)) && !(rawClass.equals(Mono.class))) {
```
rassmate, line 50: Not flux?

JonasHallFnx: I only added Mono for now because it was needed in `ParamResolvers.getResolverTargetClass` where we fetch a `Mono<T>` that is going to be a parameter to the method. My plan was to follow up with a separate PR addressing the different types of reactive implementations, to allow the use of Flux and others anywhere in the system.

utils/src/test/java/se/fortnox/reactivewizard/util/FluxTest.java
JonasHallFnx: Correct, removed.

Issue #10, opened 2017-12-01
Add checkstyle rules


JonasHallFnx:
We made a decision on spaces rather than tabs.

PR #196, opened 2020-08-10
Cannot read multidimensional lists in configmap

pontusjo:
Newlines are read as strings if a multidimensional list is added to configmap.

Example of YAML structure that dont work in configmap:
`nicheBanks:  clients:    - testclient      - "12345"`
 
Error message:
`ERRO[2020-07-08T10:19:01.954502254+02:00] v1.ConfigMap.Data: ReadString: expects " or n, but found [, error found in #10 byte of ...|clients":[{"testclient|..., bigger context ...|3.8","database.port":"1234","nichebanks.clients":[ {"testclient":["12345"]}]},"kind":"Co|...`

### DIFF
--- a/config/src/main/java/se/fortnox/reactivewizard/config/ConfigReader.java
+++ b/config/src/main/java/se/fortnox/reactivewizard/config/ConfigReader.java
@@ -56,6 +56,9 @@ public class ConfigReader {
         StringBuffer        stringBuffer = new StringBuffer();
         do {
             String value = env.get(matcher.group(1));
+            if (value != null) {
+                value = value.replace("\\n", "\n");
+            }
             matcher.appendReplacement(stringBuffer, value == null ? "" : Matcher.quoteReplacement(value));
         }
         while (matcher.find());
--- a/config/src/test/java/se/fortnox/reactivewizard/config/ConfigReaderTest.java
+++ b/config/src/test/java/se/fortnox/reactivewizard/config/ConfigReaderTest.java
@@ -159,4 +159,22 @@ public class ConfigReaderTest {
             assertThat(exception).hasRootCauseInstanceOf(MarkedYAMLException.class);
         }
     }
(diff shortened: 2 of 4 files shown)


config/src/main/java/se/fortnox/reactivewizard/config/ConfigReader.java
```diff
@@ -56,6 +56,9 @@ private static String replaceEnvPlaceholders(String config) {
         StringBuffer        stringBuffer = new StringBuffer();
         do {
             String value = env.get(matcher.group(1));
+            if (value != null) {
+                value = value.replace("\\n", "\n");
```
splitfeed, line 60: Wouldn't this cause problems when you have strings that are supposed to contain an escaped newline?
like `emailBody: "Hi\\nWelcome to thing!"`

pontusjo: Tried searching for escaped newlines but could not find any results https://git.fnox.se/plugins/servlet/search?q=%22%5C%5Cn%22

pontusjo: Do we really need to support escaped newlines inside YAML's?

tsfcode: Do we have any example of newlines being the expected outcome of reading the yaml? Since this will be breaking those configs. Otherwise if that becomes a problem we could add an escape sequence that's used for actual newlines.

splitfeed: After some discussions we've come to believe that there are no good reason for having escaped newlines like that, so we're gonna go ahead and try this out.
If this doesn't work there is a workaround in the featureswitch service that can be used.

PR #197, opened 2020-08-17
Making sure the no content fix is not called twice if it already exists

rassmate:
Making sure the NoContentFix is only added once


server/src/main/java/se/fortnox/reactivewizard/server/NoContentFixConfigurator.java
```diff
@@ -26,7 +26,7 @@ public void call(ChannelPipeline pipeline) {
         if (pipeline.get(NO_CONTENT_FIX) != null) {
             return;
         }
-        pipeline.addBefore(HttpTrafficHandler, "NoContentFix", new NoContentBodyFix());
+        pipeline.addBefore(HttpTrafficHandler, NO_CONTENT_FIX, new NoContentBodyFix());
```
rassmate: Nice catch 👍

PR #200, opened 2020-08-24
Supporting getting the meta data from the response from an api call..

rassmate:
Sending the api call through this new fullResponse method will return a wrapped object containing data, status and headers from the response.

Lets say you have a UserResource like this:

```java
interface UserResource {

    Observable<User> getUsers();
}
```

The normal way to use it is 

```java
userResource.getUsers().map(listofusers -> {
    //inspect data
    return listofusers
});

```
With this method you can
```java
HttpClient.fullResponse(userResource.getUsers()).map(responsewithdata -> {
    //inspect headers
    responsewithdata.getHeaders()
    //inspect status
    responsewithstatus.getStatus()
    //inspect data
    responsewithstatus.getData();

    return responsewithdata.getData();
});
```

### DIFF
--- a/client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
+++ b/client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
@@ -683,7 +683,7 @@ public class HttpClient implements InvocationHandler {
 
     /**
      * Class used when the full response is needed
-     * @param <T>
+     * @param <T> the type of data to be returned
      */
     public static class Response<T> {
         private final T data;
@@ -711,7 +711,7 @@ public class HttpClient implements InvocationHandler {
 
     /**
      * Internal class to help support getting the full response as return value when observable is returned
-     * @param <T>
+     * @param <T> the type of data to be returned
      */
     private static class ObservableWithResponse<T> extends Observable<T> {
 
@@ -732,7 +732,7 @@ public class HttpClient implements InvocationHandler {
 
     /**
      * Internal class to help support getting the full response as return value when Single is returned
-     * @param <T>
+     * @param <T> the type of data to be returned
      */
     private static class SingleWithResponse<T> extends Single<T> {
 
--- a/client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
+++ b/client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
@@ -187,7 +187,7 @@ public class HttpClient implements InvocationHandler {
      * from the response.
      * @param source the source observable, must be observable returned from api call
      * @param <T> the type of data that should be returned in the call
-     * @return an observable that along with the data passes the response object from netty
+     * @return an observable that along with the data passes the response meta data
      */
     public static <T> Observable<Response<T>> fullResponse(Observable<T> source) {
         if (!(source instanceof ObservableWithResponse)) {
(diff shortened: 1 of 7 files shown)


JonasHallFnx (approved):

client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
```diff
@@ -169,9 +177,35 @@ public Object invoke(Object proxy, Method method, Object[] arguments) {
         publisher = withRetry(request, flux).onErrorResume(e -> convertError(request, e));
 
         if (Single.class.isAssignableFrom(method.getReturnType())) {
-            return RxReactiveStreams.toSingle(publisher);
+            return new SingleWithResponse(RxReactiveStreams.toSingle(publisher), rawResponse);
+        }
+        return new ObservableWithResponse(RxReactiveStreams.toObservable(publisher), rawResponse);
+    }
+
+    /** Should be used with
```
jepp3, line 185: should be used with? :)

client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
```diff
@@ -169,9 +177,35 @@ public Object invoke(Object proxy, Method method, Object[] arguments) {
...
+     * @param source the source observable, must be observable returned from api call
+     * @param <T> the type of data that should be returned in the call
+     * @return an observable that along with the data passes the response object from netty
+     */
+    public static <T> Observable<Response<T>> fullResponse(Observable<T> source) {
```
jepp3, line 190: when are these used?

jepp3 (approved):

client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
rassmate: Only in tests.
Should be used in systems that needs response metadata along with the actual data

splitfeed:
Could you  update the description with an example of how this will look in the service using it? The API here is not obvious :)

rassmate:
> Could you update the description with an example of how this will look in the service using it? The API here is not obvious :)
Fixed. And you could also look at the tests for a hint on how to use it

rassmate:
When I tried it out I realized I had to make a way to easily mock these calls with headers.
So I added the class HttpClientTestUtil for that.

client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
```diff
@@ -646,4 +682,74 @@ public BeanParamProperty(Function<Object, Object> getter, Annotation[] annotatio
             this.annotations = annotations;
         }
     }
+
+    /**
+     * Class used when the full response is needed
+     * @param <T> the type of data to be returned
+     */
+    public static class Response<T> {
```
splitfeed, line 690: Is there any sane way of keeping these three classes out of HttpClient? It's already huge :)

client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
```diff
@@ -169,9 +177,37 @@ public Object invoke(Object proxy, Method method, Object[] arguments) {
...
+    public static <T> Observable<Response<T>> getFullResponse(Observable<T> source) {
+        if (!(source instanceof ObservableWithResponse)) {
+            throw new IllegalArgumentException("Must be used with observable returned from api call");
         }
-        return RxReactiveStreams.toObservable(publisher);
+
+        return source.map(data -> new Response<>(((ObservableWithResponse<T>)source).getResponse(), data));
+    }
+
+    /** Should be used with
```
splitfeed, line 200: Missing a line break

client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
rassmate: What a good suggestion. Had them here before when i tried to have them private. But not there is no need. I fix

splitfeed (approved):

PR #201, opened 2020-08-24
Redact value of Authorization header from logs

j-holmstrom:
The purpose of this change is to allow us to hide the value of specific headers that contains sensitive information.
Right now, only the Authorization header is redacted.

### DIFF
--- a/client/src/test/java/se/fortnox/reactivewizard/client/HttpClientTest.java
+++ b/client/src/test/java/se/fortnox/reactivewizard/client/HttpClientTest.java
@@ -67,6 +67,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.Date;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
@@ -612,7 +613,9 @@ public class HttpClientTest {
     public void shouldRedactAuthorizationHeaderInLogs() throws Exception {
         DisposableServer                   server       = startServer(BAD_REQUEST, "someError");
         HttpClientConfig                   config       = new HttpClientConfig("localhost:" + server.port());
-        config.setDevHeaders(Map.of("Authorization", "secretvalue"));
+        Map<String, String>                headers      = new HashMap<>();
+        headers.put("Authorization", "secretvalue");
+        config.setDevHeaders(headers);
         TestResource                       resource     = getHttpProxy(config);
 
         assertThatExceptionOfType(WebException.class)
--- a/client/src/test/java/se/fortnox/reactivewizard/client/HttpClientTest.java
+++ b/client/src/test/java/se/fortnox/reactivewizard/client/HttpClientTest.java
@@ -13,7 +13,9 @@ import io.netty.handler.ssl.util.SelfSignedCertificate;
 import org.apache.log4j.Appender;
 import org.apache.log4j.Level;
 import org.apache.log4j.LogManager;
+import org.junit.After;
 import org.junit.Assert;
+import org.junit.Before;
 import org.junit.Test;
 import org.mockito.Mockito;
 import org.reactivestreams.Publisher;
@@ -67,6 +69,7 @@ import java.util.Collections;
 import java.util.Date;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicBoolean;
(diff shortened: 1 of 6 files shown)


jepp3 (approved):

pontusjo (approved):

client/src/test/java/se/fortnox/reactivewizard/client/HttpClientTest.java
```diff
@@ -849,6 +885,8 @@ public void shouldCloseConnectionOnTimeoutDuringContentReceive() throws URISynta
     @Test
     public void shouldShouldNotGivePoolExhaustedIfServerDoesNotCloseConnection() throws URISyntaxException {
```
splitfeed, line 886: Was this somehow related to your changes, or did it resolve some previous issue?

splitfeed (approved):

jaxrs/src/main/java/se/fortnox/reactivewizard/jaxrs/RequestLogger.java
```diff
@@ -35,10 +42,43 @@ public static void headersToString(HttpServerRequest request, StringBuilder logL
...
     }
 
+    /**
+     * Returns the value of a header. If the header contains sensitive information the value will be replaced with "REDACTED".
+     *
+     * @param header The header to get the value from
+     * @return value of the header or "REDACTED" if the header contains sensitive information
+     */
+    public static String getHeaderValueOrRedact(Map.Entry<String, String> header) {
+        if ("Authorization".equalsIgnoreCase(header.getKey())) {
```
rassmate, line 59: A small optimization would be to add "Authorization" and "Redacted" strings att static strings

jaxrs/src/main/java/se/fortnox/reactivewizard/jaxrs/RequestLogger.java
```diff
@@ -35,10 +42,43 @@ public static void headersToString(HttpServerRequest request, StringBuilder logL
...
+        return header.getValue();
+    }
+
+    /**
+     * Redact sensitive information from header values. Any header that contains sensitive information will have the value replaced with "REDACTED"
+     *
+     * @param headers The headers to map
+     * @return headers with sensitive information redacted
+     */
+    public static Set<Map.Entry<String, String>> getHeaderValuesOrRedact(Map<String, String> headers) {
```
rassmate, line 71: Maybe check that headers map is not null before using it?

jaxrs/src/main/java/se/fortnox/reactivewizard/jaxrs/RequestLogger.java
```diff
@@ -22,8 +26,11 @@ public RequestLogger(Logger logger) {
      * @param logLine StringBuilder to append the header content to.
      */
     public static void headersToString(HttpServerRequest request, StringBuilder logLine) {
-        request.requestHeaders().forEach(e ->
-            logLine.append(e.getKey()).append('=').append(e.getValue()).append(' ')
+        request.requestHeaders().forEach(header ->
+            logLine.append(header.getKey())
```
rassmate, line 30: Nice formatting! 🌮

jaxrs/src/main/java/se/fortnox/reactivewizard/jaxrs/RequestLogger.java
```diff
@@ -35,10 +42,43 @@ public static void headersToString(HttpServerRequest request, StringBuilder logL
      */
     public static void headersToString(HttpServerResponse response, StringBuilder logLine) {
         response.responseHeaders().forEach(header ->
-            logLine.append(header.getKey()).append('=').append(header.getValue()).append(' ')
+            logLine.append(header.getKey())
+                .append('=')
```
rassmate, line 46: And here. ⭐

rassmate (approved):
Great work!
Only minor comments

client/src/test/java/se/fortnox/reactivewizard/client/HttpClientTest.java
j-holmstrom: This resolved an issue where this test turned off logging for my test case which made it fail.

jaxrs/src/main/java/se/fortnox/reactivewizard/jaxrs/RequestLogger.java
JonasHallFnx: From a performance point of view, I don't think it matters

JonasHallFnx (approved):

jaxrs/src/main/java/se/fortnox/reactivewizard/jaxrs/RequestLogger.java
rassmate: Will the compiler turn it into a static final? Otherwise the strings are freshly created each time and we will gain a performance boost from making it static final.

But not many calls come through here so it makes small difference.

jaxrs/src/main/java/se/fortnox/reactivewizard/jaxrs/RequestLogger.java
```diff
@@ -69,6 +71,9 @@ public static String getHeaderValueOrRedact(Map.Entry<String, String> header) {
      * @return headers with sensitive information redacted
      */
     public static Set<Map.Entry<String, String>> getHeaderValuesOrRedact(Map<String, String> headers) {
+        if(headers == null) {
+            return null;
```
rassmate, line 75: Maybe we can return emptySet() here instead. Not sure the parameter is ever null or returning null will cause trouble but better safe than sorry

PR #203, opened 2020-08-31
Added a way to wrap an Observable-/SingleWithResponse

rassmate:
Our subsystem uses it's own subclass extending HttpClient and makes the original solution not working at all times. We need a way to create new ObservableWithResponses when our internal httpclient applies some some operators on the ObservableWithResponse.

This was a way to make it possible but not too easy so that everyone would use it.

### DIFF
--- a/client/src/main/java/se/fortnox/reactivewizard/client/SingleWithResponse.java
+++ b/client/src/main/java/se/fortnox/reactivewizard/client/SingleWithResponse.java
@@ -24,4 +24,8 @@ class SingleWithResponse<T> extends Single<T> {
         }
         return httpClientResponse.get();
     }
+
+    static <T> SingleWithResponse<T> from(SingleWithResponse<T> singleWithResponse, Single<T> inner) {
+        return new SingleWithResponse<T>(inner, singleWithResponse.httpClientResponse);
+    }
 }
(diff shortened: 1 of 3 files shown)


client/src/main/java/se/fortnox/reactivewizard/client/SingleWithResponse.java
```diff
@@ -24,4 +24,8 @@ HttpClientResponse getResponse() {
         }
         return httpClientResponse.get();
     }
+
+    static <T> SingleWithResponse<T> from(SingleWithResponse<T> singleWithResponse, Single<T> inner) {
```
jepp3, line 28: these  have no defined scope, where are these going to be used?

jepp3 (commented):
Could you add a description to the PR describing why we are doing it and what problem it solves :)

JonasHallFnx (approved):

JonasHallFnx:
Nice solution!

client/src/main/java/se/fortnox/reactivewizard/client/SingleWithResponse.java
rassmate: It should be used In our subsystem but should not easily used, thus the package private scope.

jepp3 (approved):

flowertwig (approved):

splitfeed (approved):

PR #205, opened 2020-09-03
Change CollectionOptions to allow extending

rassmate:
QueryParam annotations was previously added to the CollectionOptions properties to allow for an api to extend this class to make the api definition sleek.

Extending CollectionOptions proved not fully working in our internal framework so we decided to delete CollectionOptionsWithResult and put that extra property in the CollectionOptions class instead.


api/src/main/java/se/fortnox/reactivewizard/CollectionOptions.java
```diff
@@ -69,4 +71,12 @@ public SortOrder getOrder() {
     public void setOrder(SortOrder order) {
         this.order = order;
     }
+
+    public boolean isLastRecord() {
+        return lastRecord;
+    }
+
+    public void setNotLastRecord() {
```
JonasHallFnx, line 79: If we in the future would like to make this a nullable property, then a setter taking a boolean would mean that the api would stay (almost) intact?

JonasHallFnx (approved):

splitfeed (approved):

PR #208, opened 2020-09-08
adding a method to retrieve cookie values from the response

rassmate:
Added a way to fetch cookie values to the full response

### DIFF
--- a/client/src/main/java/se/fortnox/reactivewizard/client/Response.java
+++ b/client/src/main/java/se/fortnox/reactivewizard/client/Response.java
@@ -8,6 +8,7 @@ import reactor.netty.http.client.HttpClientResponse;
 import java.util.List;
 import java.util.Map;
 
+import static java.util.Collections.emptyList;
 import static java.util.stream.Collectors.toList;
 
 /**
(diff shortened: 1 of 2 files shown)


carl-pagels (approved):

client/src/main/java/se/fortnox/reactivewizard/client/Response.java
```diff
@@ -37,7 +37,23 @@ public HttpResponseStatus getStatus() {
...
+
+    /**
+     * Fetching a cookie value
+     * @param cookieName the cookie name
+     * @return the value(s) of the cookie
+     */
+    public List<String> getCookie(String cookieName) {
+        return httpClientResponse
+            .cookies()
+            .get(cookieName)
```
flowertwig, line 53: Will get return null or an empty set if there is no cookie with that name? A test for this would be good.

flowertwig (changes requested):

JonasHallFnx (approved):

flowertwig (approved):

client/src/main/java/se/fortnox/reactivewizard/client/Response.java
```diff
@@ -37,7 +38,29 @@ public HttpResponseStatus getStatus() {
...
-        return headers.get(header);
+        return httpClientResponse
+            .responseHeaders()
+            .get(header);
+    }
+
+    /**
+     * Fetching a cookie value
+     * @param cookieName the cookie name
+     * @return the value(s) of the cookie
```
jepp3, line 49: are the values the properties of the cookie? or can a cookie have multiple values? 

Should we have it case sensetive?

rassmate: It is the values that are returned. A cookie can have multiple values.
Cookie names are case sensitive

PR #210, opened 2020-09-24
Backpressure fix

rassmate:
Fix backpressureproblem by removing explicit calls to subscriber.request with max value.

All other tests still run so I take that as a valid change without adding any new tests for this.


rassmate:
Creating new one due to branch issues

PR #211, opened 2020-09-24
removing explicit calls to subscribers request method.

rassmate:
Removing unwanted calls to subscribers request method.

All tests still passes so I take that as a good sign and will not add new tests just for this

### DIFF
--- a/metrics/src/main/java/se/fortnox/reactivewizard/metrics/MetricsOperator.java
+++ b/metrics/src/main/java/se/fortnox/reactivewizard/metrics/MetricsOperator.java
@@ -29,14 +29,11 @@ public class MetricsOperator<T> implements Publisher<T> {
 
     private Subscriber<T> createSubscriber(Subscriber<? super T> target, Timer timer) {
         return new Subscriber<T>() {
-            private Subscription subscription;
             private Timer.Context context;
             @Override
             public void onSubscribe(Subscription subscription) {
                 context = timer.time();
-                this.subscription = subscription;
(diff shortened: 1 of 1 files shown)


JonasHallFnx (approved):

metrics/src/main/java/se/fortnox/reactivewizard/metrics/MetricsOperator.java
```diff
@@ -29,22 +29,18 @@ public void subscribe(Subscriber<? super T> subscriber) {
...
     private Subscriber<T> createSubscriber(Subscriber<? super T> target, Timer timer) {
         return new Subscriber<T>() {
-            private Subscription subscription;
             private Timer.Context context;
             @Override
             public void onSubscribe(Subscription subscription) {
                 context = timer.time();
-                this.subscription = subscription;
                 target.onSubscribe(subscription);
-                this.subscription.request(Long.MAX_VALUE);
```
magnusp, line 39: Weird results over at sonarcloud but I don't think we have any tests at all covering MetricsOperator. Perhaps we should test this with an application that uses it?

magnusp (changes requested):

metrics/src/main/java/se/fortnox/reactivewizard/metrics/MetricsOperator.java
magnusp: Tested with internal application and both timers and counters were published on incoming requests

magnusp (approved):

PR #213, opened 2020-10-15
Getting all fields from ancestors as well when serializing beanparam annotate…

rassmate:
When calling other services with @Bean param annotaded parameters the http client would not serialize fields from superclasses. 
This pr will traverse the class hieararchy and and add all declared fields that should be used to the path

### DIFF
--- a/client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
+++ b/client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
@@ -7,6 +7,7 @@ import com.fasterxml.jackson.databind.JsonMappingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.ObjectReader;
 import com.fasterxml.jackson.databind.type.TypeFactory;
+import com.google.common.collect.Sets;
 import io.netty.handler.codec.http.HttpResponseStatus;
 import io.netty.handler.timeout.ReadTimeoutException;
 import org.reactivestreams.Publisher;
@@ -58,6 +59,7 @@ import java.util.ArrayList;
 import java.util.Base64;
 import java.util.Date;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
@@ -578,7 +580,7 @@ public class HttpClient implements InvocationHandler {
 
     private List<BeanParamProperty> getBeanParamGetters(Class beanParamType) {
         List<BeanParamProperty> result = new ArrayList<>();
-        for (Field field : getDeclaredFieldsFromClassAndAncestors(beanParamType, new ArrayList<>())) {
+        for (Field field : getDeclaredFieldsFromClassAndAncestors(beanParamType)) {
             Optional<Function<Object, Object>> getter = ReflectionUtil.getter(beanParamType, field.getName());
             if (getter.isPresent()) {
                 result.add(new BeanParamProperty(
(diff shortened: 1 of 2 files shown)


JonasHallFnx (approved):

client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
```diff
@@ -590,6 +590,24 @@ protected String getPath(Method method, Object[] arguments, JaxRsMeta meta) {
...
+     *
+     * @param clazz the clazz
+     * @param fields the fields found so far.
+     *
+     * @return list of fields
+     */
+    private List<Field> getDeclaredFieldsFromClassAndAncestors(Class clazz, List<Field> fields) {
+        fields.addAll(asList(clazz.getDeclaredFields()));
+
+        if (clazz.getSuperclass() != null && !Object.class.equals(clazz.getSuperclass())) {
```
jepp3, line 604: Usually the terminal condition is placed in the top, when doing recursion. In this cade it would make sens. We would then not need the negation inside the if?

A field may be declared twice? In a subclass and in the patent? Maybe a set would be better than a list?

jepp3 (approved):

client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
```diff
@@ -590,6 +590,24 @@ protected String getPath(Method method, Object[] arguments, JaxRsMeta meta) {
...
 
+    /**
+     * Recursive function getting all declared fields from the passed in class and its ancestors
+     *
+     * @param clazz the clazz
+     * @param fields the fields found so far.
+     *
+     * @return list of fields
+     */
+    private List<Field> getDeclaredFieldsFromClassAndAncestors(Class clazz, List<Field> fields) {
```
jepp3, line 601: If the method is private, then it could be static as well :)

client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
```diff
@@ -590,6 +590,24 @@ protected String getPath(Method method, Object[] arguments, JaxRsMeta meta) {
...
+        fields.addAll(asList(clazz.getDeclaredFields()));
```
jepp3, line 602: If we have apache collections, we could do a union of two sets without mutating the provided list :) return SetUtils. union(a, b) ;

jepp3:
My coments are just thoughts of improvment, nothing that you need to change :)

client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
rassmate: A set wouldn't change anything since the Field object is tied to its declaring class and a Set would still contain all fields.
And if they were the same a Field in the parent class could have the @QueryParam annotation and the field in the subclass  not so it is desirable to get all fields.

client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
rassmate: 👍

client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
```diff
@@ -590,6 +592,21 @@ protected String getPath(Method method, Object[] arguments, JaxRsMeta meta) {
         return result;
     }
 
+    /**
+     * Recursive function getting all declared fields from the passed in class and its ancestors
+     *
+     * @param clazz the clazz fetching fields from
+     * @return list of fields
+     */
+    private static Set<Field> getDeclaredFieldsFromClassAndAncestors(Class clazz) {
```
jepp3, line 601: Wow! :)

jepp3 (approved):

PR #215, opened 2020-10-20
Added possibility to enable gzip for certain mime-types.

j-holmstrom:
Gzip will be enabled if we set enableGzip = true in ServerConfig.
Compression will only be enabled for the following mime-types.

- text/plain
- application/xml
- text/css 
- application/x-javascript
- application/json

We will compress response payloads exceeding 1000 bytes. Chunked responses are not compressed.

Resources with stream annotation will not be compressed ( and should not be ).

### DIFF
--- a/server/src/main/java/se/fortnox/reactivewizard/server/RwServer.java
+++ b/server/src/main/java/se/fortnox/reactivewizard/server/RwServer.java
@@ -31,7 +31,7 @@ import static reactor.netty.channel.BootstrapHandlers.updateConfiguration;
 public class RwServer extends Thread {
 
     private static final Logger            LOG                         = LoggerFactory.getLogger(RwServer.class);
-    private static final Set<String>       COMPRESS_CONTENT_TYPES      = new HashSet<>(asList("text/plain", "application/xml", "text/css application/x-javascript", "application/json"));
+    private static final Set<String>       COMPRESS_CONTENT_TYPES      = new HashSet<>(asList("text/plain", "application/xml", "text/css", "application/x-javascript", "application/json"));
     private static final int               COMPRESSION_THRESHOLD_BYTES = 1000;
     private final        ServerConfig      config;
     private final        ConnectionCounter connectionCounter;
--- a/server/src/main/java/se/fortnox/reactivewizard/server/RwServer.java
+++ b/server/src/main/java/se/fortnox/reactivewizard/server/RwServer.java
@@ -1,9 +1,12 @@
 package se.fortnox.reactivewizard.server;
 
+import io.netty.handler.codec.http.HttpHeaderNames;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import reactor.netty.DisposableServer;
 import reactor.netty.http.server.HttpServer;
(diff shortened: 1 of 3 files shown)


server/src/main/java/se/fortnox/reactivewizard/server/RwServer.java
```diff
@@ -78,6 +91,20 @@ private static HttpServer createHttpServer(ServerConfig config, LoopResources lo
                 .maxHeaderSize(config.getMaxHeaderSize()));
     }
 
+    private static BiPredicate<HttpServerRequest, HttpServerResponse> compressableType() {
+        return (request, response) -> {
+            String contentType = response.responseHeaders()
+                .get(HttpHeaderNames.CONTENT_TYPE);
+            return Optional.ofNullable(contentType)
+                .map(COMPRESS_CONTENT_TYPES::contains)
```
JonasHallFnx, line 99: How come we don't compress all types?

j-holmstrom: The cpu cost to compress already compressed files like images/pdfs is larger than the network gain from compressing them.

server/src/main/java/se/fortnox/reactivewizard/server/RwServer.java
```diff
@@ -78,6 +91,20 @@ private static HttpServer createHttpServer(ServerConfig config, LoopResources lo
                 .maxHeaderSize(config.getMaxHeaderSize()));
     }
 
+    private static BiPredicate<HttpServerRequest, HttpServerResponse> compressableType() {
+        return (request, response) -> {
+            String contentType = response.responseHeaders()
```
magnusp, line 96: See if we can use https://netty.io/4.1/api/io/netty/handler/codec/http/HttpUtil.html#getMimeType-java.lang.CharSequence- instead. Also make sure that mimetype is always lowercased when matching.

server/src/test/java/se/fortnox/reactivewizard/server/RwServerGzipTest.java
```diff
@@ -0,0 +1,126 @@
...
+import static org.apache.commons.lang.RandomStringUtils.randomAlphabetic;
+import static org.assertj.core.api.Assertions.assertThat;
+
+@RunWith(MockitoJUnitRunner.class)
+public class RwServerGzipTest {
+    private RwServer rwServer;
+
+    @Test
+    public void shouldCompressAllowedContentTypes() {
+        Arrays.asList("text/plain", "application/xml", "text/css", "application/x-javascript", "application/json")
```
magnusp, line 27: Extend with examples that include directives: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Type

splitfeed (approved):

magnusp (changes requested):

jepp3 (approved):

pontusjo (approved):

splitfeed (approved):

magnusp (approved):

JonasHallFnx (approved):

PR #219, opened 2020-10-26
Mark headers as sensitive on resources for outgoing calls

claesloof:
This change allows you to mark specified headers as sensitive on resources for outgoing calls, redacting the header values from logging.

Example: `HttpClient.markHeaderAsSensitive(someResource, "cookie")`

### DIFF
--- a/client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
+++ b/client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
@@ -57,6 +57,7 @@ import java.time.temporal.ChronoUnit;
 import java.time.temporal.TemporalUnit;
 import java.util.ArrayList;
 import java.util.Base64;
+import java.util.Comparator;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -64,9 +65,11 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
+import java.util.TreeSet;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Consumer;
 import java.util.function.Function;
 
 import static io.netty.handler.codec.http.HttpMethod.POST;
@@ -75,7 +78,7 @@ import static io.netty.handler.codec.http.HttpResponseStatus.INTERNAL_SERVER_ERR
 import static java.lang.String.format;
 import static java.util.Arrays.asList;
 import static java.util.Collections.emptySet;
-import static java.util.Collections.singletonList;
+import static java.util.Collections.singleton;
 import static javax.ws.rs.core.HttpHeaders.CONTENT_TYPE;
 import static reactor.core.Exceptions.isRetryExhausted;
 import static rx.Observable.fromCallable;
@@ -97,7 +100,7 @@ public class HttpClient implements InvocationHandler {
     private final   Map<Method, JaxRsMeta>                            jaxRsMetaMap     = new ConcurrentHashMap<>();
     private         int                                               timeout          = 10;
     private         TemporalUnit                                      timeoutUnit      = ChronoUnit.SECONDS;
-    private         List<String>                                      sensitiveHeaders = new ArrayList<>();
+    private         Set<String>                                       sensitiveHeaders = new TreeSet<>(Comparator.comparing(String::toLowerCase));
     private final   Duration                                          retryDuration;
 
     @Inject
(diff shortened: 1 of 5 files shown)


splitfeed (approved):

magnusp (approved):

jepp3 (approved):

client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
```diff
@@ -125,16 +127,34 @@ public static void setTimeout(Object proxy, int timeout, ChronoUnit timeoutUnit)
...
             }
         }
     }
 
+    public static void markHeaderAsSensitive(Object proxy, String header) {
+        markHeadersAsSensitive(proxy, singletonList(header));
+    }
+
+    public static void markHeadersAsSensitive(Object proxy, List<String> headers) {
+        if (Proxy.isProxyClass(proxy.getClass())) {
```
rassmate, line 140: What do you think about creating a small method that does this check and accepts a lambda as a parameter where we can actually do the mutation of the HttpClient object

With this way we can reuse that method both here and in the setTimeout method.
Something like this:

private static void ifHttpClientDo(Object proxy, Consumer<HttpClient> consumer) {
    if (Proxy.isProxyClass(proxy.getClass()){
Object handler = Proxy.getInvocationHandler(proxy);
if (handler instanceof HttpClient) {
consumer.accept((HttpClient)handler);
}

} 
}

jaxrs/src/main/java/se/fortnox/reactivewizard/jaxrs/RequestLogger.java
```diff
@@ -57,10 +63,23 @@ public static void headersToString(HttpServerResponse response, StringBuilder lo
...
+     * @return value of the header or "REDACTED" if the header contains sensitive information
+     */
+    public static String getHeaderValueOrRedact(Map.Entry<String, String> header, List<String> sensitiveHeaders) {
         if (header == null) {
             return null;
-        } else if ("Authorization".equalsIgnoreCase(header.getKey())) {
-            return "REDACTED";
+        } else if (AUTHORIZATION_HEADER.equalsIgnoreCase(header.getKey())) {
+            return REDACTED;
+        } else if (sensitiveHeaders.stream().anyMatch(header.getKey()::equalsIgnoreCase)) {
```
JonasHallFnx, line 81: This will run on every header of every log, so would it be possible to have a `Set` instead so that we get fast lookups here?

jaxrs/src/main/java/se/fortnox/reactivewizard/ExceptionHandler.java
```diff
@@ -68,7 +68,7 @@ public ExceptionHandler() {
         } else if (throwable instanceof WebException) {
             webException = (WebException)throwable;
         } else if (throwable instanceof ClosedChannelException) {
-            LOG.debug("ClosedChannelException: " + request.method() + " " + request.uri(), throwable);
+            LOG.debug("ClosedChannelException: {} {}", request.method(), request.uri(), throwable);
```
rassmate, line 71: 🌟

JonasHallFnx (changes requested):
Change to a Set for better lookup performance

jaxrs/src/main/java/se/fortnox/reactivewizard/jaxrs/RequestLogger.java
claesloof: Good idea, I'm on it :)

client/src/test/java/se/fortnox/reactivewizard/client/HttpClientTest.java
```diff
@@ -717,6 +717,35 @@ public void shouldRedactAuthorizationHeaderInLogsAndExceptionMessage() throws Ex
...
+        DisposableServer                   server       = startServer(BAD_REQUEST, "someError");
+        HttpClientConfig                   config       = new HttpClientConfig("localhost:" + server.port());
+        Map<String, String>                headers      = new HashMap<>();
+        headers.put("Cookie", "pepperidge farm");
+        config.setDevHeaders(headers);
+        TestResource                       resource     = getHttpProxy(config);
+
+        HttpClient.markHeaderAsSensitive(resource, "cookie");
+
+        assertThatExceptionOfType(WebException.class)
```
rassmate, line 731: FYI we now have:

ObservableAssertions in the framework where we can do observable assertions.
ObservableAssertions.assertThatExceptionOfType(NoSuchElementException.class)
			.isEmittedBy(observableWithoutToBlockingSingle)

rassmate (changes requested):
Nice pr, i would like some small refactorings in HttpClient

client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
claesloof: Done! Looks much neater now. Thanks!

rassmate (approved):

client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
```diff
@@ -92,49 +96,67 @@
...
     public static void setTimeout(Object proxy, int timeout, ChronoUnit timeoutUnit) {
+       ifHttpClientDo(proxy, handler -> handler.setTimeout(timeout, timeoutUnit));
+    }
+
+    public static void markHeaderAsSensitive(Object proxy, String header) {
+        markHeadersAsSensitive(proxy, singleton(header));
+    }
+
+    public static void markHeadersAsSensitive(Object proxy, Set<String> headers) {
+        ifHttpClientDo(proxy, handler -> handler.setSensitiveHeaders(headers));
```
rassmate, line 138: Nitpicking here but maybe here and in setTimout name the lambda parameter httpClient to make it clearer that it is in fact an httpClient instance coming in.

JonasHallFnx (approved):

PR #222, opened 2020-10-27
when client aborts, the exceptionhandler should debug-log it but not …

rassmate:
…throw exception since it is so commonly occuring.

### DIFF
--- a/jaxrs/src/main/java/se/fortnox/reactivewizard/ExceptionHandler.java
+++ b/jaxrs/src/main/java/se/fortnox/reactivewizard/ExceptionHandler.java
@@ -9,6 +9,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
+import reactor.netty.channel.AbortedException;
 import reactor.netty.http.server.HttpServerRequest;
 import reactor.netty.http.server.HttpServerResponse;
 import rx.exceptions.CompositeException;
@@ -67,7 +68,7 @@ public class ExceptionHandler {
             webException = new WebException(HttpResponseStatus.BAD_REQUEST, "invalidjson", throwable.getMessage());
         } else if (throwable instanceof WebException) {
             webException = (WebException)throwable;
-        } else if (throwable instanceof ClosedChannelException) {
+        } else if (throwable instanceof ClosedChannelException || throwable instanceof AbortedException) {
             LOG.debug("ClosedChannelException: {} {}", request.method(), request.uri(), throwable);
             return Flux.empty();
         } else {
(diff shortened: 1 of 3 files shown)


jaxrs/src/main/java/se/fortnox/reactivewizard/ExceptionHandler.java
```diff
@@ -67,7 +68,7 @@ public ExceptionHandler() {
             webException = new WebException(HttpResponseStatus.BAD_REQUEST, "invalidjson", throwable.getMessage());
         } else if (throwable instanceof WebException) {
             webException = (WebException)throwable;
-        } else if (throwable instanceof ClosedChannelException) {
+        } else if (throwable instanceof ClosedChannelException || throwable instanceof AbortedException) {
             LOG.debug("ClosedChannelException: {} {}", request.method(), request.uri(), throwable);
```
jepp3, line 72: Maybe change the debug message to something like: "Inbound connection has been closed"? 

What is the difference between closedChannel and abortedexception?

jepp3 (approved):

server/src/test/java/se/fortnox/reactivewizard/server/AbortedRequestHandlingTest.java
```diff
@@ -0,0 +1,91 @@
...
+
+        RwServer      rwServer                   = null;
+        AtomicBoolean serverFinishedWithoutError = new AtomicBoolean(false);
+
+        try {
+            JaxRsRequestHandler jaxRsRequestHandler = new JaxRsRequestHandler(new Object[]{new TestResourceImpl()}, new JaxRsResourceFactory(), new ExceptionHandler(), false);
+
+            rwServer = server((httpServerRequest, httpServerResponse) -> {
+                httpServerRequest.withConnection(connection -> {
+                    connection.channel().close();
```
JonasHallFnx, line 48: This is closing the connection on the server side, but the problem that we see is that the client closes the connection. Perhaps it results in the same error, but I think it is a different scenario. Maybe the client closing is harder to simulate due to http keepalive and stuff.

Have you tried just createing a server with a delay in the response, and have a client with a really short timeout set? That might result in the client closing the connection before the server writes the response.

rassmate: Yes, I tried that at first but didn't succeed. I will give it another try.

Yes, we close the connection at the server side but it will generate the same type of exception that is thrown when the client closes the connection. 
I think who closes the connection is not the important thing but rather that the client connection is closed when trying to send data.

JonasHallFnx (approved):

jaxrs/src/main/java/se/fortnox/reactivewizard/ExceptionHandler.java
rassmate: Fixed. 
With RxNetty we got the raw ClosedChannelException.
Now it comes in a AbortedException as the cause

tsfcode (approved):

jepp3 (approved):

PR #223, opened 2020-10-29
Bugfix/graceful shutdown

jepp3:
Registered a channel group so that graceful shutdown can be made.
Coop with @magnusp, @j-holmstrom

### DIFF
--- a/server/src/main/java/se/fortnox/reactivewizard/server/ServerConfig.java
+++ b/server/src/main/java/se/fortnox/reactivewizard/server/ServerConfig.java
@@ -59,7 +59,7 @@ public class ServerConfig {
         return shutdownTimeoutSeconds;
     }
 
-    public void setShutdownTimeout(int shutdownTimeoutSeconds) {
+    public void setShutdownTimeoutMs(int shutdownTimeoutSeconds) {
         this.shutdownTimeoutSeconds = shutdownTimeoutSeconds;
     }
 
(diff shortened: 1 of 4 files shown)


j-holmstrom (approved):

server/src/main/java/se/fortnox/reactivewizard/server/ServerConfig.java
```diff
@@ -59,7 +59,7 @@ public int getShutdownTimeoutSeconds() {
         return shutdownTimeoutSeconds;
     }
 
-    public void setShutdownTimeoutMs(int shutdownTimeoutSeconds) {
+    public void setShutdownTimeout(int shutdownTimeoutSeconds) {
```
splitfeed, line 62: Is it the method or property name that is mapped to the config value? Just to make sure this isn't a breaking change :)

jepp3: Yes this will be a breaking change i think. I will revert it. it might cause problems for people when updating to newer version. But the name is indeed not very good :)

jepp3: I will revert the change, so that no breaking change is introduced.

splitfeed (approved):

JonasHallFnx (approved):

pontusjo (approved):

tsfcode (approved):

PR #226, opened 2020-11-05
Adding WaitingTestUtils to the framework as a way to handle timing is an issue

rassmate:
As a response to the problem:
As a developer I want to write a unit test that asserts that a condition is met in a certain time.

### DIFF
--- a/test/src/main/java/se/fortnox/reactivewizard/test/WaitingTestUtils.java
+++ b/test/src/main/java/se/fortnox/reactivewizard/test/WaitingTestUtils.java
@@ -6,7 +6,7 @@ import rx.Subscription;
 
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
-import java.util.function.Supplier;
+import java.util.function.BooleanSupplier;
 
 import static java.util.concurrent.TimeUnit.MILLISECONDS;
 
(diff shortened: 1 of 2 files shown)


test/src/main/java/se/fortnox/reactivewizard/test/WaitingTestUtils.java
```diff
@@ -0,0 +1,63 @@
...
+ */
+public abstract class WaitingTestUtils {
+
+    private static final String DEFAULT_ERROR_MESSAGE = "Condition not met within time";
+    private static final int DEFAULT_WAITING_SECONDS = 5;
+    private static final TimeUnit DEFAULT_WAITING_TIME_UNIT = TimeUnit.SECONDS;
+
+    private WaitingTestUtils() {}
+
+    public static void assertConditionIsTrueWithin5Seconds(Supplier<Boolean> condition) {
```
tsfcode: assertConditionIsTrueWithinDefaultTimeout mabey, otherwise if we want to change the default timeout, these functions should change name, breaking tests.

test/src/test/java/se/fortnox/reactivewizard/test/WaitingTestUtilsTest.java
```diff
@@ -0,0 +1,77 @@
...
+
+        assertThat(assertionError).isNotNull();
+        assertThat(assertionError.getMessage()).isEqualTo(errorMessage);
+    }
+
+    @Test
+    public void shouldFailTestIfConditionIsNotMetWithinTime() {
+        Error assertionError = null;
+        try {
+            WaitingTestUtils.assertConditionIsTrueWithinTime(3000, TimeUnit.MILLISECONDS, () -> false);
```
tsfcode, line 46: Nitpick: adding 3 seconds for a test feels like a bit much, the functionallity should be able to be tested with 50ms?

rassmate: The testutil checks every 100ms so I will raise it to the least possible value

tsfcode (approved):

jepp3 (approved):

JonasHallFnx (approved):

j-holmstrom (approved):

PR #228, opened 2020-11-24
make the httpclient more resilient to failures when connecting.

rassmate:
A single failure should not cause the health recorder to report unhealthy. Adding a config for this, and also adding a config for how long a connection can remain idle before thrown away.

### DIFF
--- a/client/src/main/java/se/fortnox/reactivewizard/client/HttpClientConfig.java
+++ b/client/src/main/java/se/fortnox/reactivewizard/client/HttpClientConfig.java
@@ -52,6 +52,9 @@ public class HttpClientConfig {
     @JsonProperty("validateCertificates")
     private boolean isValidateCertificates = true;
 
+    private long connectionMaxIdleTimeInMs         = TimeUnit.MILLISECONDS.convert(10, MINUTES);
+    private int  numberOfConnectionFailuresAllowed = 10;
+
     private BasicAuthConfig basicAuth;
 
     public HttpClientConfig() {
@@ -197,5 +200,21 @@ public class HttpClientConfig {
 
     public void setPoolAcquireTimeoutMs(int poolAcquireTimeoutMs) {
         this.poolAcquireTimeoutMs = poolAcquireTimeoutMs;
+    }
+
+    public long getConnectionMaxIdleTimeInMs() {
+        return connectionMaxIdleTimeInMs;
+    }
+
+    public void setConnectionMaxIdleTimeInMs(long connectionMaxIdleTimeInMs) {
+        this.connectionMaxIdleTimeInMs = connectionMaxIdleTimeInMs;
+    }
+
+    public int getNumberOfConnectionFailuresAllowed() {
+        return numberOfConnectionFailuresAllowed;
+    }
+
+    public void setNumberOfConnectionFailuresAllowed(int numberOfConnectionFailuresAllowed) {
+        this.numberOfConnectionFailuresAllowed = numberOfConnectionFailuresAllowed;
     }
 }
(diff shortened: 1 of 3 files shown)


JonasHallFnx (approved):

client/src/main/java/se/fortnox/reactivewizard/client/ReactorRxClientProvider.java
```diff
@@ -58,8 +62,13 @@ private HttpClient buildClient(InetSocketAddress socketAddress) {
                 })
             )
             .port(config.getPort())
-            .doOnRequest((httpClientRequest, connection) -> healthRecorder.logStatus(connectionProvider, true))
-            .doOnError((httpClientRequest, throwable) -> healthRecorder.logStatus(connectionProvider, false), (httpClientResponse, throwable) -> { })
+            .doOnRequest((httpClientRequest, connection) -> {
```
magnusp, line 65: With doOnRequest we reset the errorcounter before the request is sent, just after it has been prepared. Shouldn't we reset the counter after the response has been received fully but before handing it off to the handler?

jepp3 (approved):

client/src/main/java/se/fortnox/reactivewizard/client/ReactorRxClientProvider.java
rassmate: Yes, but the connection is established so I think it is fine.

client/src/main/java/se/fortnox/reactivewizard/client/HttpClientConfig.java
```diff
@@ -52,6 +52,9 @@
     @JsonProperty("validateCertificates")
     private boolean isValidateCertificates = true;
 
+    private long connectionMaxIdleTimeInMs         = TimeUnit.MILLISECONDS.convert(10, MINUTES);
```
splitfeed, line 55: so what does this actually affect? That a connection may be inactive for 10 minutes before the pod marks itself as unhealthy?

client/src/main/java/se/fortnox/reactivewizard/client/ReactorRxClientProvider.java
```diff
@@ -58,8 +62,13 @@ private HttpClient buildClient(InetSocketAddress socketAddress) {
...
+                errorCount.set(0);
+                healthRecorder.logStatus(connectionProvider, true);
+            })
+            .doOnError((httpClientRequest, throwable) -> {
+                healthRecorder.logStatus(connectionProvider, errorCount.incrementAndGet() < config.getNumberOfConnectionFailuresAllowed());
```
splitfeed, line 70: Won't this be false when there are 10 errors even if 10 are configured as allowed?

Your test also says "First nine should not cause the client to be unhealthy", but that conflicts with the configuration saying "Number of connection failures allowed". 
If I configure to allow 10 errors then 11 should be considered unhealthy, right?

splitfeed (changes requested):

client/src/main/java/se/fortnox/reactivewizard/client/HttpClientConfig.java
splitfeed: Does that mean that a low traffic service that doesn't get more than one call every 10 minutes will get killed due to inactivity?

rassmate: This is just a way to clean up idle connections, and has nothing to do with unhealthyness.
So if a connection is created, it will be usable for 10 minutes. If noone uses it for 10 minutes, it will will be removed from the pool and when the next request comes it will create a new connection to be put in the pool.

It has nothing with the unhealthyness to do. 
I introduced again, we had it before, it because the tests would not behave the way I wanted otherwise.
And before we had a cleanup interval for 10 minutes.

client/src/main/java/se/fortnox/reactivewizard/client/ReactorRxClientProvider.java
rassmate: Correct

rassmate: I fix

tsfcode (approved):

PR #230, opened 2021-01-18
returning the raw response body string if it does not start with quotes

rassmate:


JonasHallFnx (approved):

magnusp (approved):

jepp3 (approved):

client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
```diff
@@ -528,6 +530,10 @@ private void setHeaderParams(RequestBuilder request, Method method, Object[] arg
             return Mono.empty();
         }
 
+        if (String.class.equals(type) && !string.startsWith(QOUTES) && !"null".equalsIgnoreCase(string)) {
```
jepp3, line 533: qoutes -> quotes

jepp3: maybe im thinking wrong here. but should it not be: 

if( is String and starts with quote ) ?

splitfeed: > qoutes -> quotes

I actually think it should be just quote, since it only contains one quotation mark  :)

client/src/test/java/se/fortnox/reactivewizard/client/HttpClientTest.java
```diff
@@ -961,6 +962,18 @@ public void shouldDeserializeVoidResult() {
         server.disposeNow();
     }
 
+    @Test
+    public void shouldDeserializeSimpleStringWithoutJsonQuotes() {
```
splitfeed, line 966: Seems like you are missing a test case for when the response is `"null"` . Are there tests that cover strings that are quoted, so that is validated to be working?

splitfeed (changes requested):

client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
rassmate: If the string starts with quotes we need to run it through the objectMapper

rassmate: I will change spelling and remove an S

The null test is already implemented in RestClientFactoryTest

client/src/test/java/se/fortnox/reactivewizard/client/HttpClientTest.java
rassmate: The "null" test is implemented in RestClientFactoryTest.shouldDeserializeNullToEmpty()

jepp3 (approved):

splitfeed (approved):

pontusjo (approved):

PR #233, opened 2021-03-10
Added five second wait before server shutdown for improved Kubernetes interoperability

flowertwig:
Removed tests for graceful shutdown since they did not test what was intended.
Removed LoopResources from RwServer since it was never used.
Upped reactor-bom to SR17.

### DIFF
--- a/server/src/main/java/se/fortnox/reactivewizard/server/RwServer.java
+++ b/server/src/main/java/se/fortnox/reactivewizard/server/RwServer.java
@@ -41,7 +41,6 @@ public class RwServer extends Thread {
         "application/x-javascript",
         "application/json"
     ));
-    private static final long SECONDS_WAIT_BEFORE_SHUTDOWN = 5;
 
     private final ServerConfig config;
     private final ConnectionCounter connectionCounter;
@@ -137,11 +136,12 @@ public class RwServer extends Thread {
     }
 
     static void shutdownHook(ServerConfig config, DisposableServer server, ConnectionCounter connectionCounter) {
-        LOG.info("Shutdown requested. Waiting {} seconds before commencing.", SECONDS_WAIT_BEFORE_SHUTDOWN);
+        LOG.info("Shutdown requested. Waiting {} seconds before commencing.", config.getShutdownDelaySeconds());
         try {
-            Thread.sleep(SECONDS_WAIT_BEFORE_SHUTDOWN * 1000);
+            Thread.sleep(config.getShutdownDelaySeconds() * 1000);
         } catch (InterruptedException e) {
             LOG.error("Interrupted while waiting for shutdown to commence.", e);
+            Thread.currentThread().interrupt();
         }
         LOG.info("Shutdown commencing. Will wait up to {} seconds for ongoing requests to complete.", config.getShutdownTimeoutSeconds());
         int elapsedSeconds = measureElapsedSeconds(() ->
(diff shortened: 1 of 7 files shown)


server/src/main/java/se/fortnox/reactivewizard/server/RwServer.java
```diff
@@ -145,16 +136,22 @@ public static void registerShutdownDependency(Runnable blockShutdownUntil) {
         RwServer.blockShutdownUntil = blockShutdownUntil;
     }
 
-    static void shutdownHook(ServerConfig config, DisposableServer server, LoopResources loopResources, ConnectionCounter connectionCounter) {
-        LOG.info("Shutdown requested. Will wait up to {} seconds...", config.getShutdownTimeoutSeconds());
+    static void shutdownHook(ServerConfig config, DisposableServer server, ConnectionCounter connectionCounter) {
+        LOG.info("Shutdown requested. Waiting {} seconds before commencing.", SECONDS_WAIT_BEFORE_SHUTDOWN);
+        try {
+            Thread.sleep(SECONDS_WAIT_BEFORE_SHUTDOWN * 1000);
```
jepp3, line 142: since the code already is blocking this is okay we think and we are on our way to shutdown the service.

jepp3 (approved):

server/src/main/java/se/fortnox/reactivewizard/server/RwServer.java
```diff
@@ -33,38 +33,33 @@
...
+    private static final Logger LOG = LoggerFactory.getLogger(RwServer.class);
+    private static final int COMPRESSION_THRESHOLD_BYTES = 1000;
+    private static final Set<String> COMPRESSIBLE_MIME_TYPES = new HashSet<>(asList(
         "text/plain",
         "application/xml",
         "text/css",
         "application/x-javascript",
         "application/json"
     ));
+    private static final long SECONDS_WAIT_BEFORE_SHUTDOWN = 5;
```
rassmate, line 44: Should we add this to the ServerConfig instead so that we can make configuration changes if it would be necessary ?

server/src/main/java/se/fortnox/reactivewizard/server/RwServer.java
```diff
@@ -33,38 +32,32 @@
...
+    private static Runnable blockShutdownUntil;
 
     @Inject
-    public RwServer(ServerConfig config, CompositeRequestHandler compositeRequestHandler, ConnectionCounter connectionCounter) {
-        this(config, compositeRequestHandler, connectionCounter, null);
-    }
-
-    RwServer(ServerConfig config, CompositeRequestHandler compositeRequestHandler, ConnectionCounter connectionCounter, LoopResources loopResources) {
-        this(config, connectionCounter, createHttpServer(config, loopResources), compositeRequestHandler, loopResources);
+    RwServer(ServerConfig config, CompositeRequestHandler compositeRequestHandler, ConnectionCounter connectionCounter) {
```
JonasHallFnx, line 51: Removing public here intentionally? Isn't this a breaking change? No problems for Guice to create this?

flowertwig: Guice seems fine with this, but it wasn't intentional, so I'll fix.

rassmate (approved):

PR #235, opened 2021-05-06
Check for duplicate resource paths and missing pathparam annotations …

j-holmstrom:
…on startup.

Allow easy creation of startup checks by implementing the StartupCheck interface.

relates to PLAT-1331


JonasHallFnx:
Just curious; how come you need to register a "startup check" - can't you just check whatever needs to be checked in the constructor? Is there some state that is not ready at that time?

jaxrs/src/main/java/se/fortnox/reactivewizard/jaxrs/startupchecks/StartupCheckConfig.java
```diff
@@ -0,0 +1,17 @@
+package se.fortnox.reactivewizard.jaxrs.startupchecks;
+
+import se.fortnox.reactivewizard.config.Config;
+
+@Config("startupcheck")
+public class StartupCheckConfig {
+    private boolean failOnError = true;
```
JonasHallFnx, line 7: How come you need this to be configurable? Since it is bound to the code rather than to the environment - can you use an annotation or similar on the resource method to signal to the checker that it is ok with colliding paths? That would remove the dependency on the config module and would mean that you don't need to keep the configs up to date with the framework revision.

jaxrs/src/main/java/se/fortnox/reactivewizard/jaxrs/startupchecks/CheckForDuplicatePaths.java
```diff
@@ -0,0 +1,159 @@
...
+    @Inject
+    public CheckForDuplicatePaths(StartupCheckConfig startupCheckConfig) {
+        this.startupCheckConfig = startupCheckConfig;
+    }
+
+    @Override
+    public void check(List<JaxRsResource> resources) {
+        resources.forEach(jaxRsResource -> {
+            final Optional<JaxRsResource> duplicate = resources.stream()
+                .filter(res -> res != jaxRsResource && pathParamCollides(res, jaxRsResource))
```
JonasHallFnx, line 40: You will be comparing both (a, b) and (b, a) here. Not a big problem for production, but perhaps this code adds up to some seconds in your everyday work when you are restarting services locally and also in tests?

jaxrs/src/main/java/se/fortnox/reactivewizard/jaxrs/startupchecks/CheckForDuplicatePaths.java
```diff
@@ -0,0 +1,159 @@
...
+            final Type thisPathParamType = getTypeOfParam(thisMatcher.group(1), jaxRsResource1.getResourceMethod());
+            final Type thatPathParamType = getTypeOfParam(thatMatcher.group(1), jaxRsResource2.getResourceMethod());
+
+            if (!Objects.equals(thisPathParamType, thatPathParamType)) {
+                pathParamCollision = true;
+            }
+        }
+
+        //Different verbs causes no collision
+        if (!jaxRsResource1.getHttpMethod().equals(jaxRsResource2.getHttpMethod())) {
```
JonasHallFnx, line 76: This check is very cheap, moving it up would save some CPU from running the code above with the matchers.

jaxrs/src/main/java/se/fortnox/reactivewizard/jaxrs/startupchecks/StartupCheckConfig.java
j-holmstrom: Good idea. 

We only wanted the config so that we could ignore the error in the case there is some case we haven't thought about.
We plan to remove the config once we see there are no issues.

jaxrs/src/main/java/se/fortnox/reactivewizard/jaxrs/CheckForDuplicatePaths.java
```diff
@@ -0,0 +1,160 @@
...
+    @Inject
+    public CheckForDuplicatePaths(StartupCheckConfig startupCheckConfig, JaxRsRequestHandler jaxRsRequestHandler) {
+        this.startupCheckConfig = startupCheckConfig;
+        final List<JaxRsResource> resources = jaxRsRequestHandler.getResources().getResources();
+
+        check(resources);
+    }
+
+    void check(List<JaxRsResource> resources) {
+        for (int i = 0; i < resources.size() - 1; i++) {
```
jepp3, line 38: a lot of I and J. Maybe a iterator is better to use? 

```
  void check(List<JaxRsResource> resources) {
        Iterator<JaxRsResource> iterator = resources.iterator();
        if(iterator.hasNext()) {
            JaxRsResource current = iterator.next();
            iterator.forEachRemaining(suspect -> checkIfColliding(current, suspect));
        }
    }

    private void checkIfColliding(JaxRsResource current, JaxRsResource other) {
        if (collidingPaths(current, other)) {
            String errorMessage = String.format("%s duplicates %s on path %s",
                current.getResourceMethod(),
                other.getResourceMethod(),
                other.getPath()
            );
            if (startupCheckConfig.isFailOnError()) {
                throw new IllegalStateException(errorMessage);
            } else {
                LOG.warn(errorMessage);
            }
        }
    }
```

jaxrs/src/main/java/se/fortnox/reactivewizard/jaxrs/CheckForDuplicatePaths.java
```diff
@@ -0,0 +1,160 @@
...
+import java.lang.reflect.Parameter;
+import java.lang.reflect.Type;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Objects;
+import java.util.Set;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class CheckForDuplicatePaths extends StartupCheck {
```
jepp3, line 23: Duplicate/colliding maybe stick with one? or conflicting?

jaxrs/src/main/java/se/fortnox/reactivewizard/jaxrs/CheckForDuplicatePaths.java
```diff
@@ -0,0 +1,160 @@
...
+        if (jaxRsResource2 == null) {
+            return false;
+        }
+
+        //Different verbs causes no collision
+        if (!jaxRsResource1.getHttpMethod().equals(jaxRsResource2.getHttpMethod())) {
+            return false;
+        }
+
+        final Matcher thisMatcher = PATH_PARAM_PATTERN.matcher(jaxRsResource1.getPath());
```
jepp3, line 68: could we have the same name while comparing, making it easier to follow in the code.  resourceToTest, otherResource -> jaxRsResource1, jaxRsResource2  -> this , that :)

jaxrs/src/main/java/se/fortnox/reactivewizard/jaxrs/CheckForDuplicatePaths.java
```diff
@@ -0,0 +1,160 @@
...
+
+        final String thisPath          = createPathPattern(jaxRsResource1.getPath()).toString();
+        final String jaxRsResourcePath = createPathPattern(jaxRsResource2.getPath()).toString();
+
+        //The pattern differs somehow
+        if (!thisPath.equalsIgnoreCase(jaxRsResourcePath)) {
+            return false;
+        }
+
+        return pathParamCollision || checkOtherParameters(jaxRsResource1, jaxRsResource2);
```
jepp3, line 89: maybe we can avoid having pathParamCollision variable at all? 

```
  final Matcher thisMatcher = PATH_PARAM_PATTERN.matcher(jaxRsResource1.getPath());
        final Matcher thatMatcher = PATH_PARAM_PATTERN.matcher(jaxRsResource2.getPath());

        final String thisPath          = createPathPattern(jaxRsResource1.getPath()).toString();
        final String jaxRsResourcePath = createPathPattern(jaxRsResource2.getPath()).toString();

        //The pattern differs somehow
        if (!thisPath.equalsIgnoreCase(jaxRsResourcePath)) {
            return false;
        }

        while (thisMatcher.find() && thatMatcher.find()) {
            final Type thisPathParamType = getTypeOfParam(thisMatcher.group(1), jaxRsResource1.getResourceMethod());
            final Type thatPathParamType = getTypeOfParam(thatMatcher.group(1), jaxRsResource2.getResourceMethod());

            if (!Objects.equals(thisPathParamType, thatPathParamType)) {
                return true;
            }
        }
        
        return checkOtherParameters(jaxRsResource1, jaxRsResource2);
```

jaxrs/src/main/java/se/fortnox/reactivewizard/jaxrs/CheckForDuplicatePaths.java
```diff
@@ -0,0 +1,160 @@
...
+                    } else {
+                        LOG.warn(errorMessage);
+                    }
+                }
+            }
+        }
+    }
+
+    public static boolean pathParamCollides(JaxRsResource jaxRsResource1, JaxRsResource jaxRsResource2) {
+        if (jaxRsResource2 == null) {
```
jepp3, line 59: this if statement should not be needed? we should never be able to get into this state if the loop works as intended?

jaxrs/src/main/java/se/fortnox/reactivewizard/jaxrs/CheckForDuplicatePaths.java
```diff
@@ -0,0 +1,160 @@
...
+            thisMethod = resource1.getInstanceMethod();
+        }
+
+        Method thatMethod = ReflectionUtil.getOverriddenMethod(resource2.getInstanceMethod());
+        if (thatMethod == null) {
+            thatMethod = resource2.getInstanceMethod();
+        }
+
+        Set<String> thisParams = extractParameterAnnotations(thisMethod);
+        Set<String> thatParams = extractParameterAnnotations(thatMethod);
```
jepp3, line 122: smart! :+1:

jaxrs/src/test/java/se/fortnox/reactivewizard/jaxrs/JaxRsResourceTest.java
```diff
@@ -742,6 +843,21 @@ public void shouldAcceptBeanParamInherited() {
         assertThat(get(service, "/test/acceptsBeanParamInherited?name=foo&age=3&items=1,2&inherited=YES").getOutp()).isEqualTo("\"foo - 3 2 - YES\"");
     }
 
+    @Test
+    public void test() {
```
jepp3, line 847: hum :)

jaxrs/src/main/java/se/fortnox/reactivewizard/jaxrs/CheckForDuplicatePaths.java
```diff
@@ -0,0 +1,160 @@
...
+
+        final Matcher thisMatcher = PATH_PARAM_PATTERN.matcher(jaxRsResource1.getPath());
+        final Matcher thatMatcher = PATH_PARAM_PATTERN.matcher(jaxRsResource2.getPath());
+
+        boolean pathParamCollision = false;
+        while (thisMatcher.find() && thatMatcher.find()) {
+            final Type thisPathParamType = getTypeOfParam(thisMatcher.group(1), jaxRsResource1.getResourceMethod());
+            final Type thatPathParamType = getTypeOfParam(thatMatcher.group(1), jaxRsResource2.getResourceMethod());
+
+            if (!Objects.equals(thisPathParamType, thatPathParamType)) {
```
jepp3, line 76: /internalapi/stuff/{myvariable}
/internalapi/stuff/{myvariable2}

myvariable2 = string
myvariable1 = date

if types are not the same, is it ok then? :confused:

j-holmstrom: No, you are right.
The type check should be removed!

jepp3 (approved):

jaxrs/src/main/java/se/fortnox/reactivewizard/jaxrs/CheckForCollidingPaths.java
```diff
@@ -0,0 +1,139 @@
...
+                    CookieParam param = (CookieParam)annotation;
+                    params.add("CookieParam: " + param.value());
+                }
+            }
+        }
+        return params;
+    }
+
+    private static Pattern createPathPattern(String path) {
+        // Paths with custom regex, like this: {myvar:myregex}
```
splitfeed, line 127: I did not even know this was a thing :D

splitfeed (approved):
This looks great, I'm sure several hours of wasteful debugging will be saved with this 👍

rassmate (approved):

j-holmstrom:
Closing this PR in favor of #237 that does the same thing without all the extra infrastructure.

PR #237, opened 2021-05-12
Checking resources for colliding paths and missing path params on startup.

j-holmstrom:
Added check for colliding resource paths on startup.
Added check for missing PathParam annotations on startup.

Closing PR #235 in favor of this.

### DIFF
--- a/jaxrs/src/main/java/se/fortnox/reactivewizard/jaxrs/SuppressPathCollision.java
+++ b/jaxrs/src/main/java/se/fortnox/reactivewizard/jaxrs/SuppressPathCollision.java
@@ -1,0 +1,11 @@
+package se.fortnox.reactivewizard.jaxrs;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.METHOD)
+public @interface SuppressPathCollision {
+}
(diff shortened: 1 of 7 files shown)


jaxrs/src/main/java/se/fortnox/reactivewizard/jaxrs/SuppressPathCollision.java
```diff
@@ -0,0 +1,11 @@
+package se.fortnox.reactivewizard.jaxrs;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.METHOD)
+public @interface SuppressPathCollision {
```
JonasHallFnx, line 10: There is already a java.lang.SuppressWarnings annotation that could be used I guess. Then the users of the annotation would not have to add jaxrs module to classpath (it is not part of dependencies in api-modules). If you for some reason need to create a new annotation, then perhaps it is nice to have it in module `jaxrs-api`? Or perhaps this should just go on the impl-class and then jaxrs-module will be on CP anyway...

j-holmstrom: The SuppressWarnings does not have runtime retention so I had to create my own annotation.
Moving it to the jaxrs-api mdule sounds good though.

JonasHallFnx (approved):

jepp3 (approved):

PR #239, opened 2021-05-20
Allow streaming of response values by using the @Stream annotation on resource interfaces

TommyPersson:
Example declarations (from `HttpClientTest.java`)
```java
interface TestResource {
        @GET
        @Path("string-stream")
        @Produces("text/plain")
        @Stream
        Observable<String> streamingStrings();

        @GET
        @Path("pojo-stream")
        @Stream
        Observable<Pojo> streamingPojos();
}
```

Breaking changes:
* `HttpClient.parseResponse` now returns `Flux<?>` rather than `Mono<Object>`
* `HttpClient.deserialize` now returns `Flux<Object>` rather than `Mono<Object>`

Notes:
* This change also allows the client to parse streamed JSON objects even without the annotation, but the objects will only be outputted after the entire response has been read. Previously only the first object in the stream was made available.
* ~~JSON stream parsing assumes that the object boundaries precisely match the HTTP chunks. (This appears to be the case when streamed from an RW server.)~~
  * This was fixed in a38479fccb944a0ebdf4995b08b885886d3e41e8 with the `JsonTokenBufferer`
* It is not possible to stream "primitive" json values such as numbers since there are no delimiters in the streamed response.

### DIFF
--- a/client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
+++ b/client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
@@ -23,6 +23,7 @@ import rx.Single;
 import se.fortnox.reactivewizard.jaxrs.ByteBufCollector;
 import se.fortnox.reactivewizard.jaxrs.FieldError;
 import se.fortnox.reactivewizard.jaxrs.JaxRsMeta;
+import se.fortnox.reactivewizard.jaxrs.Stream;
 import se.fortnox.reactivewizard.jaxrs.WebException;
 import se.fortnox.reactivewizard.metrics.HealthRecorder;
 import se.fortnox.reactivewizard.metrics.PublisherMetrics;
@@ -80,7 +81,6 @@ import static java.util.Collections.emptySet;
 import static java.util.Collections.singleton;
 import static javax.ws.rs.core.HttpHeaders.CONTENT_TYPE;
 import static reactor.core.Exceptions.isRetryExhausted;
-import static reactor.core.publisher.Mono.just;
 import static rx.Observable.fromCallable;
 import static se.fortnox.reactivewizard.jaxrs.RequestLogger.getHeaderValuesOrRedact;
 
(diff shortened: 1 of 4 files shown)


JonasHallFnx:
I have another branch, https://github.com/FortnoxAB/reactive-wizard/tree/feature/flux-mono-return-types, which is WIP, that is ment to solve this by using Flux/Mono and Flowable/Single to denote the single/multiple returntype. I think that is a good way going forward to allow this problem to be solved for both client and server, which is not dependent on our custom annotation @Stream. That annotation is currently ment to be put on implementation rather than on interface, to avoid leaking this to api/clients.

I also think that assuming that chunks matches objects will break when there is a proxy or loadbalancer in between, which is the case when not running locally.

I would love to discuss the best way to go forward with this. Good job!

client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
```diff
@@ -519,25 +537,25 @@ private void setHeaderParams(RequestBuilder request, Method method, Object[] arg
...
+            return Flux.just(string);
         }
 
         try {
             JavaType     javaType = TypeFactory.defaultInstance().constructType(type);
             ObjectReader reader   = objectMapper.readerFor(javaType);
-            Object       value    = reader.readValue(string);
-            return Mono.justOrEmpty(value);
+            List<Object> value    = reader.readValues(string).readAll();
+            return Flux.fromArray(value.toArray());
```
rassmate, line 558: Flux.fromIterable could be used to avoid putting it into an array

But Even nicer would have been to avoid the reader.readValues(string.readAll()) as a blocking call outside the observbable as well when we are at it.

What do you think about:

return Flux.fromStream(() -> {
            try {
                return reader.readValues(string).readAll().stream();
            } catch (IOException e) {
                throw new RuntimeException(e);
            }
        });

client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
```diff
@@ -177,28 +177,22 @@ public Object invoke(Object proxy, Method method, Object[] arguments) {
...
+
+            return parseResponse(method, request, rwHttpClientResponse);
+        });
+
         publisher = measure(request, publisher);
 
-        //End of publisher
         Flux<?> flux = Flux.from(publisher);
-        flux = flux.timeout(Duration.of(timeout, timeoutUnit));
+        flux      = flux.timeout(Duration.of(timeout, timeoutUnit));
```
rassmate, line 195: Not related to your code change but putting this duration into a field instead would have been nice

rassmate:
Agree with @JonasHallFnx  that perhaps we should have a discussion about how to take this forward.

PR #240, opened 2021-05-21
Feature/flux mono return types

JonasHallFnx:
Adds support for using `Flux` and `Mono` as return types for resources and dao-interfaces. `Flux` types are serialized as json-arrays over the wire, which means that you can now send a stream without buffering the whole result in memory and instead use backpressure between server and client. Using a `Flux` can now replace usage of the `@Stream` annotation.

`Flux` and `Observable` can co-exist in the same resources for easier migration.

Other types of reactive-streams implementations (like RxJava2) can now be added more easily as well, but there is some more work to be done to get it pluggable.

This implementation should not affect the existing implementation, and it should be fully backwards compatible. The new Flux return type should be tested in small scale first, there might be stuff in the json-deserializing that can consume memory in an unpredicted way. There are some alternatives to the current implementation.


client/src/main/java/se/fortnox/reactivewizard/client/JsonArrayDeserializer.java
```diff
@@ -0,0 +1,69 @@
...
+        reader   = objectMapper.readerFor(javaType);
+        try {
+            parser = objectMapper.getFactory().setCodec(objectMapper).createNonBlockingByteArrayParser();
+            inputFeeder = (ByteArrayFeeder) parser.getNonBlockingInputFeeder();
+        } catch (Exception e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    public Publisher<?> process(byte[] buf) {
```
jepp3, line 38: will it work with nested objects?

JonasHallFnx: I belive so, but will add a test to prove it :-)

TommyPersson: It would also be nice to have tests for compatibility with the "old-style" streaming method. (Sequential objects without delimiters)

JonasHallFnx: Like this?
```
{"name":"A"}{"name":"B"}
```
Not sure we want to support that in the client. Wouldn't it be strange if that did not give an error for bad json?

TommyPersson: Perphaps switch behavior on a "legacy" marker annotation? I would like to be able to call currently existing streaming resources without much hassle.

rassmate (changes requested):
Great Job!!
Really nice with the reading streaming responses stuff.

Some unused imports and some questions for you.
Reviewed with @jepp3 @flowertwig @j-holmstrom

client/src/main/java/se/fortnox/reactivewizard/client/JsonArrayDeserializer.java
rassmate: The old @Stream annotated methods is an Observable<SMTH> and that will still work i guess but no streaming when calling. 
The @Stream annotation was only intended for the server side.

If you want to make use of the new stuff before the old @Stream annotated endpoint is migrated to Flux<SMTH> then you could easily create your own interface in your own codebase mimicing the one from the maven dependency and just change the return type in the local interface.

TommyPersson: @rassmate Sure, that's exactly what I would like to do, but I don't know of a way to declare such an interface if the new `Flux`-resources are unable to parse the current `@Stream`-responses.

rassmate: No, my bad, thought it might work out of the box but it doesn't due to everything you written here. Just a slow learner...

client/src/main/java/se/fortnox/reactivewizard/client/JsonArrayDeserializer.java
```diff
@@ -0,0 +1,69 @@
...
+        try {
+            inputFeeder.feedInput(buf, 0, buf.length);
```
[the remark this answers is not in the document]
JonasHallFnx: I don't see why that would be good...? It would mean that the buf would be allocated for a longer time and could not be GC'd as soon as it can now.

I am also considering putting the ByteBufFlux here instead for better memory performance, and then it must be consumed within the call, and you cannot do a defer.

dao/src/main/java/se/fortnox/reactivewizard/db/DbProxy.java
```diff
@@ -119,7 +121,9 @@ public Object invoke(Object proxy, Method method, Object[] args) throws Throwabl
                 databaseConfig);
             statementFactories.put(method, observableStatementFactory);
         }
-        return observableStatementFactory.create(args, connectionProvider);
+        Observable<Object> resultObservable = observableStatementFactory.create(args, connectionProvider);
```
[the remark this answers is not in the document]
JonasHallFnx: Doh! I guess I need a revisit on this one. It is kind of similar to the wrappers in the client. Perhaps we could have a generic wrapper type to avoid creating multiple wrappers. Will see if that works.

jaxrs/src/main/java/se/fortnox/reactivewizard/jaxrs/response/JaxRsStreamingResult.java
```diff
@@ -9,14 +9,19 @@
...
+    public JaxRsStreamingResult(Flux<T> output, HttpResponseStatus responseStatus, Func1<Flux<T>, Flux<byte[]>> serializer, Map<String, String> headers) {
         super(output, responseStatus, serializer, headers);
     }
 
     @Override
     public Publisher<Void> write(HttpServerResponse response) {
-        response.status(responseStatus);
-        headers.forEach(response::addHeader);
-        return response.sendByteArray(output.map(serializer::call));
+        return output.switchOnFirst((signal, outputBuffered) -> {
```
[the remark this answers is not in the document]
JonasHallFnx: Well, this is one of the more experimental parts of this PR. It seems to work, but we should keep an eye on errors due to this operator.

client/src/main/java/se/fortnox/reactivewizard/client/JsonArrayDeserializer.java
JonasHallFnx: Well, current implementations with `@Stream` has the annotation on the impl and not on the api, so I don't see how I can detect this on the client side.

I could change the server side of `@Stream` to emit valid json, but that would be a "breaking" change. Most usage of `@Stream` is for text/csv anyway, I have never used it for json, so that might be a way to go? @TommyPersson 

After this PR is merged, you could just change or add a new endpoint with a Flux return type to use the new client, would that work for you? It can co-exist with existing endpoints in the same class.

TommyPersson: There is an internal service that streams json and it's what mostly motivated the other PR. But I can accept waiting for that service to be updated with new proper streaming behavior.

client/src/main/java/se/fortnox/reactivewizard/client/JsonArrayDeserializer.java
```diff
@@ -0,0 +1,69 @@
...
+
+            List<Object> items = null;
+            Object item = null;
+            JsonToken token;
+            while ((token = parser.nextToken()) != JsonToken.NOT_AVAILABLE) {
+                if (token == JsonToken.START_ARRAY || token == JsonToken.END_ARRAY) {
```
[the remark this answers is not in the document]
JonasHallFnx: Flux<List> is now supported and will be serialized as nested arrays.

dao/src/main/java/se/fortnox/reactivewizard/db/DbProxy.java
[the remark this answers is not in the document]
JonasHallFnx: This is now solved, but will be refactored in the future.

rassmate (approved):

TommyPersson (approved):

client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
```diff
@@ -307,12 +354,12 @@ private String createBasicAuthString() {
...
     }
 
-    protected <T> Publisher<T> measure(RequestBuilder fullRequest, Publisher<T> output) {
+    protected <T> Mono<T> measure(RequestBuilder fullRequest, Mono<T> output) {
         return PublisherMetrics.get("OUT_res:" + fullRequest.getKey()).measure(output);
     }
 
-    protected <T> Flux<T> withRetry(RequestBuilder fullReq, Flux<T> response) {
-        return response.retryWhen(Retry.backoff(config.getRetryCount(), this.retryDuration).filter(throwable -> {
+    protected Retry withRetry(RequestBuilder fullReq) {
```
JonasHallFnx, line 361: restore signature

client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
```diff
@@ -413,24 +460,32 @@ protected boolean isBodyArg(@SuppressWarnings("unused") Class<?> cls, Annotation
...
-            HttpClient.DetailedError detailedError      = getDetailedError(data, detailedErrorCause);
-            String                   reasonPhrase       = detailedError.hasReason() ? detailedError.reason() : clientResponse.status().reasonPhrase();
-            HttpResponseStatus responseStatus = new HttpResponseStatus(clientResponse.status()
-                .code(),
-                reasonPhrase);
-
-            throw new WebException(responseStatus, detailedError, false);
-        }
-        return data;
+    protected Mono<Response<Flux<?>>> handleError(RequestBuilder request, RwHttpClientResponse response) {
```
JonasHallFnx, line 463: check usage

client/src/main/java/se/fortnox/reactivewizard/client/JsonArrayDeserializer.java
```diff
@@ -0,0 +1,93 @@
...
+                        if (item == null) {
+                            continue;
+                        }
+                        tokenBuffer = null;
+                        if (items != null) {
+                            items.add(item);
+                        }
+                    }
+                } catch (Exception e) {
+                    // Expected when reading fragments
```
JonasHallFnx, line 78: document what exception is expected, change catch?

client/src/test/java/se/fortnox/reactivewizard/client/FluxClientTest.java
```diff
@@ -0,0 +1,375 @@
...
+import java.util.List;
+import java.util.concurrent.atomic.AtomicInteger;
+
+import static java.util.Arrays.asList;
+import static javax.ws.rs.core.HttpHeaders.CONTENT_TYPE;
+import static javax.ws.rs.core.MediaType.APPLICATION_JSON;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.*;
+
+public class FluxClientTest {
```
JonasHallFnx, line 30: Test with non-json, but with json header

JonasHallFnx:
Master was merged into this branch.

Because a transaction will no longer actually subscribe to the underlying Observables/Fluxes, any code that needs `onTransactionCompleted` will now have to `getDecoration` and register a listener using `onTransactionCompleted` on the `StatementContext` rather than using `keepDecoration` and `doOnCompleted`, or as before, casting to `DaoObservable` and registering with `doOnTransactionCompleted`.

rassmate (changes requested):
Minor change request but just for the test to be accurate

dao/src/test/java/se/fortnox/reactivewizard/db/DaoTransactionsTest.java
```diff
@@ -72,25 +74,86 @@ public void shouldRunTwoQueriesInOneTransaction() throws SQLException {
...
-            .onTerminate(() -> cbExecuted[2] = true);
+        final boolean[] cbExecuted   = {false, false, false, false};
+        Observable<Integer> daoObsWithCb = dao.updateSuccess();
+
+        Optional<StatementContext> decoration = ReactiveDecorator.getDecoration(daoObsWithCb);
+        assertThat(decoration).isPresent();
+
+        decoration.get().onTransactionCompleted(() -> cbExecuted[0] = true);
+
+        ReactiveDecorator.keepDecoration(daoObsWithCb, obs->{
```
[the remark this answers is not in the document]
JonasHallFnx: Done.

rassmate (approved):

PR #243, opened 2021-05-31
Added a convenient method to "stream" or rather fetch data in chunks …

jepp3:
…from the database. Can be good if we want to reduce memory footprint.

Relates to: SER-6945

### DIFF
--- a/dao/src/main/java/se/fortnox/reactivewizard/db/DbChunk.java
+++ b/dao/src/main/java/se/fortnox/reactivewizard/db/DbChunk.java
@@ -1,0 +1,55 @@
+package se.fortnox.reactivewizard.db;
+
+import rx.Observable;
+import se.fortnox.reactivewizard.CollectionOptions;
+
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.function.Function;
+
+import static rx.Observable.defer;
+
+public abstract class DbChunk {
+
+    private static final int DEFAULT_CHUNK_SIZE = 500;
+
+    /**
+     * Provides a convenient way to select smaller chunks of data at the time from the database. Useful when we want to reduce the memory footprint while
+     * fetching large amout of data from the database.
+     * <p>
+     * Combine with Flux as return type or use the @Stream annotation on the resource method to stream the result to the client.
+     * <p>
+     * Example:
+     * <pre>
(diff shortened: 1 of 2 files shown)


dao/src/main/java/se/fortnox/reactivewizard/db/DbChunk.java
```diff
@@ -0,0 +1,55 @@
...
+        return defer(() -> {
+            CollectionOptions collectionOptions = new CollectionOptions();
+            collectionOptions.setOffset(currentOffset.get());
+            collectionOptions.setLimit(chunkSize);
+            return daoCall.apply(collectionOptions).toList();
+        })
+            .doOnNext((ignore) -> currentOffset.addAndGet(chunkSize))
+            .repeat()
+            .takeUntil(result -> result.size() < chunkSize)
+            .flatMapIterable(e -> e);
```
JonasHallFnx, line 43: concatMapIterable, otherwise backpressure will trigger multiple dao calls concurrently.

dao/src/main/java/se/fortnox/reactivewizard/db/DbChunk.java
```diff
@@ -0,0 +1,55 @@
...
+     *
+     * @param daoCall   A function that calls the desired dao with the supplied {@link CollectionOptions}. The collection options must be passed to the dao.
+     * @param chunkSize size of the chunks to fetch
+     * @return A observable that selects the data with the provided dao in chunks.
+     */
+    public static <T> Observable<T> inChunks(Function<CollectionOptions, Observable<T>> daoCall, int chunkSize) {
+        AtomicInteger currentOffset = new AtomicInteger();
+        return defer(() -> {
+            CollectionOptions collectionOptions = new CollectionOptions();
+            collectionOptions.setOffset(currentOffset.get());
```
JonasHallFnx, line 36: Working with offset is often a suboptimal way of "chunking" since it requires sorting, and often the database will have to process the whole resultset for each chunk anyway. I have some examples of filters where a PK is used to fetch chunks instead, which would be alot better for the db.

jepp3: forgot that aspect. I will need to look if i can use the index instead.

PR #245, opened 2021-06-09
Ignore duplicates when checking for colliding paths

flowertwig:


jepp3 (approved):

jaxrs/src/main/java/se/fortnox/reactivewizard/jaxrs/CheckForCollidingPaths.java
```diff
@@ -128,4 +129,8 @@ private static Pattern createPathPattern(String path) {
         return Pattern.compile(path);
     }
 
+    private static <T> Predicate<T> distinctByKey(Function<? super T, ?> keyExtractor) {
```
jepp3, line 132: I do think it would be easier to read if not in specified as a function

flowertwig: You mean inline it? I think distinctByKey says what it does.

rassmate (approved):

jaxrs/src/main/java/se/fortnox/reactivewizard/jaxrs/CheckForCollidingPaths.java
jepp3: I did not see that the method actually kept state  and works as a filter with state. In that case its easier with a function as here. The method is highly generic,  The T could be swapped out for the class we are working on and the variable t could be called resource to enhance readability, if the intent is not to create a framework method that requires that flexibility

jepp3: Its just a note. We dont need to change, but the statething and the generic part together makes it a bit harder to see what is what

PR #250, opened 2021-08-24
Feature/flux mono return types fixes

rassmate:


magnusp:
Take a look at the variants of then(), thenEmpty() and thenMany() on Flux which I think will cover alot of what is done in FirstThenFlux.java. Mono also has then() etc.

JonasHallFnx (approved):

PR #255, opened 2021-09-27
README files for all the modules

rassmate:
relates to NOJIRA

### DIFF
--- a/config/README.md
+++ b/config/README.md
@@ -2,12 +2,10 @@
 Enable configuration support and the possibility to `@Inject` configuration classes where needed.
 
 # How
-Configuration provided in a yml file is read, and the properties are mapped to configuration classes annotated with the `@Config`
-annotation. Configuration file is pointed out by the last argument passed when starting the `Main.class`
-There is also support for putting environment variables into the configs by setting the config value to an environment variable name 
-embraced with double curly braces.
-It will fetch the value from the environment by using the name you provide in the configuration file. 
-Useful when running the application in a container environment
+Configuration provided in a yml file is read, and the properties are mapped to configuration classes annotated with the `@Config` annotation. Configuration file is pointed out by the last argument passed when starting the `Main.class`
+There is also support for putting environment variables into the configs by setting the config value to an environment variable name embraced with double curly braces.
+It will fetch the value from the environment by using the name you provide in the configuration file.
+Useful when running the application in a container environment.
 
 ```
 export ENVIRONMENT_VARIABLE_NAME=http://host:port/path
--- a/config/README.md
+++ b/config/README.md
@@ -1,0 +1,29 @@
+# Purpose of module
+Enable configuration support and the possibility to `@Inject` configuration classes where needed.
+
+# How
+Configuration provided in a yml file is read, and the properties are mapped to configuration classes annotated with the `@Config`
+annotation. Configuration file is pointed out by the last argument passed when starting the `Main.class`
+There is also support for putting environment variables into the configs by setting the config value to an environment variable name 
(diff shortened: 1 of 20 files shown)


config/README.md
```diff
@@ -0,0 +1,29 @@
+# Purpose of module
+Enable configuration support and the possibility to `@Inject` configuration classes where needed.
+
+# How
+Configuration provided in a yml file is read, and the properties are mapped to configuration classes annotated with the `@Config`
+annotation. Configuration file is pointed out by the last argument passed when starting the `Main.class`
+There is also support for putting environment variables into the configs by setting the config value to an environment variable name 
+embraced with double curly braces.
+It will fetch the value from the environment by using the name you provide in the configuration file. 
+Useful when running the application in a container environment
```
JonasHallFnx, line 10: Missing trailing dot. No newlines needed above as the sentences are about the same thing?

dao/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+`ParameterizedQuery` - a representation of the statement that will run against the db. Capable of creating PreparedStatement to be used in the jdbc connection.
+
+`DaoTransactions` - a way to execute db-statements in an isolated transaction where _all_ statements must succeed to be committed
+
+`@Query` - the annotation denoting a method should run a select statement.
+
+`@Update` - the annotation denoting a method should run an insert/update/delete statement.
+
+# Extra important info regarding this module
```
JonasHallFnx, line 25: Good! Also, we can mention that one should use DaoTransactions any time that you need to run more than one CUD-operation, not only for transactional benefits but also for performance. Running multiple single queries (even selects) should be done using concatMap instead of flatMap, as the concurrency would otherwise saturate the connectionpool and block other requests from reaching the database. Running queries with concat will be very fast even if they are not concurrent, as one single connection will be reused and the associated caches in the database backend will often benefit from this.

Transactions in RW are special because they do not allow the developer to block the database-connection, they are "one-shot". Therefore, in order to design a parent-relation structure with foreign-keys, you need to either use UUIDs to generate all keys in code, allowing you to add all inserts to a single one-shot transaction, or if you need an ordered key, make a select to determine the next key and use the optimistic-locking pattern to let the database block concurrent insterts by failing on primary key conflict, which can be easily retried with Observables.

dbmigrate/README.md
```diff
@@ -0,0 +1,15 @@
+# Purpose of module
+Providing the possibility to set up a database migrations using `Liquibase`
+
+#How
+Reacts to command line arguments (See `LiquibaseAutoBindModule` for a complete list)
+
+Scans the classpath for files named `migrations.xml` and then uses `Liquibase` to run all migrations.xml files in an undefined order.
```
JonasHallFnx, line 7: I think there is some special stuff with in our fatjar config that merges multiple migrations.xml-files. That might be worth mentioning. Without it I suspect that maven shade will just replace one file with another.

logging/README.md
```diff
@@ -0,0 +1,6 @@
+# Purpose
+Initialize logging (done by blitz4j) through log4j.properties and through configuration in configuration file mapped to key `logging`
```
JonasHallFnx, line 2: Maybe mention that slf4j is used as api. Fyi, I have a local branch in which I am replacing blitz4j/log4j1 with log4j2, keeping the slf4j api.

JonasHallFnx (changes requested):
Very good job!

splitfeed (approved):

tsfcode (approved):

JonasHallFnx (approved):

PR #261, opened 2021-11-15
Allow systems to redact sensitive headers from logs in incoming serve…

rassmate:
…r requests.

Errors in incoming requests could print sensitive information to the logs.
This code will make it possible per service to specify what headers are sensitive and should be redacted.

### DIFF
--- a/jaxrs/src/main/java/se/fortnox/reactivewizard/ExceptionHandler.java
+++ b/jaxrs/src/main/java/se/fortnox/reactivewizard/ExceptionHandler.java
@@ -114,7 +114,7 @@ public class ExceptionHandler {
             msg
                 .append(header.getKey())
                 .append('=')
-                .append(localGetHeaderValueOrRedact(header))
+                .append(getHeaderValue(header))
                 .append(' ')
         );
         return msg.toString();
@@ -129,7 +129,7 @@ public class ExceptionHandler {
      * @return the string that should be logged for this particular header in case of an Exception.
      *
      */
-    protected String localGetHeaderValueOrRedact(Map.Entry<String, String> header) {
+    protected String getHeaderValue(Map.Entry<String, String> header) {
         return getHeaderValueOrRedact(header);
     }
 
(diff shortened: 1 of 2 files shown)


jaxrs/src/main/java/se/fortnox/reactivewizard/ExceptionHandler.java
```diff
@@ -21,15 +21,18 @@
...
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
 
 import static se.fortnox.reactivewizard.jaxrs.RequestLogger.getHeaderValueOrRedact;
 
 /**
  * Handles exceptions and writes errors to the response and the log.
  */
 public class ExceptionHandler {
```
JonasHallFnx, line 33: Just a thought to make this easier to use; Seems to me like the list of sensitive headers would fit nicely in `/server/src/main/java/se/fortnox/reactivewizard/server/ServerConfig.java`, perhaps it would be possible to make a subclass of `ExceptionHandler` in the `server` module and override `getHeaderValueOrRedact` in tht subclass, taking the `ServerConfig` as a constructor parameter?

j-holmstrom:
Maybe we should support redacting only specific values in the Cookie header?
It feels like completely redacting all cookie values in a service will make it harder to debug.

jaxrs/src/test/java/se/fortnox/reactivewizard/ExceptionHandlerTest.java
```diff
@@ -56,6 +57,25 @@ public void shouldRedactSensitiveHeaders(){
...
+    @Test
+    public void shouldRedactSensitiveHeadersSpecified(){
+
+        ExceptionHandler.addSensitiveHeader("OtherHeader");
+
+        try {
+            MockHttpServerRequest request = new MockHttpServerRequest("/path");
+            request.requestHeaders()
+                .add("Authorization", "secret")
+                .add("OtherHeader", "notasecret");
```
JonasHallFnx, line 69: `also_a_secret`?

JonasHallFnx (approved):

PR #262, opened 2021-11-24
Rewrite creation of transactions

mickej:
Instead of creating multiple observables, now one is created that executes the transaction.

It should be backwards compatible. I deprecated the createTransaction-methods, and when we are sure that no one uses them we can rewrite the transaction-stuff totally and get rid of those cross-references between objects and make a more clean implementation.

Commiting a transaction with 24.000 inserts takes around 20 seconds, 100.000 took more than 15 minutes.
This change gets it down to 3 seconds for 24k and around 20-30 seconds for 100k.

And the problem was that it creates so many observables. RxJava simply had too much to do. It creates the observables sequentially.

### DIFF
--- a/dao/src/main/java/se/fortnox/reactivewizard/db/ObservableStatementFactory.java
+++ b/dao/src/main/java/se/fortnox/reactivewizard/db/ObservableStatementFactory.java
@@ -59,15 +59,15 @@ public class ObservableStatementFactory {
 
     public Observable<Object> create(Object[] args, ConnectionProvider connectionProvider) {
         AtomicReference<TransactionStatement> transactionHolder = new AtomicReference<>();
+        Statement dbStatement = statementFactory.create(args);
 
         Observable<Object> result = Observable.unsafeCreate(subscription -> {
             try {
-                Statement dbStatement = statementFactory.create(args, subscription);
+                dbStatement.setSubscriber(subscription);
 
                 TransactionStatement transactionStatement = transactionHolder.get();
                 if (transactionStatement != null) {
                     Transaction transaction = transactionStatement.getTransaction();
-                    transaction.setConnectionProvider(connectionProvider);
                     transactionStatement.markStatementSubscribed(dbStatement);
                     if (transaction.isAllSubscribed()) {
                         scheduleWorker(subscription, transaction::execute);
@@ -94,7 +94,7 @@ public class ObservableStatementFactory {
         result = metrics.measure(result, time -> logSlowQuery(transactionHolder.get(), time, args));
         result = result.onBackpressureBuffer(RECORD_BUFFER_SIZE);
 
-        return new DaoObservable<>(result, transactionHolder);
+        return new DaoObservable<>(result, transactionHolder, dbStatement);
     }
 
     private void scheduleWorker(Subscriber<?> subscription, Action0 action) {
--- a/dao/src/main/java/se/fortnox/reactivewizard/db/ObservableStatementFactory.java
+++ b/dao/src/main/java/se/fortnox/reactivewizard/db/ObservableStatementFactory.java
@@ -11,15 +11,13 @@ import se.fortnox.reactivewizard.db.paging.PagingOutput;
 import se.fortnox.reactivewizard.db.statement.DbStatementFactory;
 import se.fortnox.reactivewizard.db.statement.Statement;
 import se.fortnox.reactivewizard.db.transactions.DaoObservable;
-import se.fortnox.reactivewizard.db.transactions.Transaction;
-import se.fortnox.reactivewizard.db.transactions.TransactionStatement;
 import se.fortnox.reactivewizard.metrics.Metrics;
 import se.fortnox.reactivewizard.util.DebugUtil;
 
 import java.sql.Connection;
 import java.sql.SQLException;
-import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Function;
+import java.util.function.Supplier;
 
 import static java.lang.String.format;
 import static rx.Observable.error;
@@ -58,23 +56,13 @@ public class ObservableStatementFactory {
     }
 
     public Observable<Object> create(Object[] args, ConnectionProvider connectionProvider) {
-        AtomicReference<TransactionStatement> transactionHolder = new AtomicReference<>();
-        Statement dbStatement = statementFactory.create(args);
-
+        Supplier<Statement> statementSupplier = () -> statementFactory.create(args);
         Observable<Object> result = Observable.unsafeCreate(subscription -> {
             try {
+                Statement dbStatement = statementSupplier.get();
                 dbStatement.setSubscriber(subscription);
 
-                TransactionStatement transactionStatement = transactionHolder.get();
-                if (transactionStatement != null) {
-                    Transaction transaction = transactionStatement.getTransaction();
-                    transactionStatement.markStatementSubscribed(dbStatement);
-                    if (transaction.isAllSubscribed()) {
-                        scheduleWorker(subscription, transaction::execute);
-                    }
-                } else {
-                    scheduleWorker(subscription, () -> executeStatement(dbStatement, connectionProvider));
-                }
+                scheduleWorker(subscription, () -> executeStatement(dbStatement, connectionProvider));
             } catch (Exception e) {
                 if (!subscription.isUnsubscribed()) {
                     subscription.onError(e);
@@ -91,10 +79,10 @@ public class ObservableStatementFactory {
         }
 
         result = pagingOutput.apply(result, args);
-        result = metrics.measure(result, time -> logSlowQuery(transactionHolder.get(), time, args));
+        result = metrics.measure(result, this::logSlowQuery);
         result = result.onBackpressureBuffer(RECORD_BUFFER_SIZE);
 
-        return new DaoObservable<>(result, transactionHolder, dbStatement);
+        return new DaoObservable<>(result, statementSupplier);
     }
 
     private void scheduleWorker(Subscriber<?> subscription, Action0 action) {
@@ -112,8 +100,8 @@ public class ObservableStatementFactory {
         });
     }
 
-    private void logSlowQuery(TransactionStatement transactionStatement, long time, Object[] args) {
-        if (transactionStatement == null && time > config.getSlowQueryLogThreshold()) {
+    private void logSlowQuery(long time) {
+        if (time > config.getSlowQueryLogThreshold()) {
             LOG.warn(format("Slow query: %s\ntime: %d", statementFactory, time));
         }
     }
--- a/dao/src/main/java/se/fortnox/reactivewizard/db/transactions/DaoObservable.java
+++ b/dao/src/main/java/se/fortnox/reactivewizard/db/transactions/DaoObservable.java
@@ -10,12 +10,10 @@ import java.util.function.Supplier;
 public class DaoObservable<T> extends Observable<T> {
     private final Observable<T> result;
     private final Supplier<Statement> statementSupplier;
-    private Action0 onTransactionCompleted;
+    private final Action0 onTransactionCompleted;
 
     public DaoObservable(Observable<T> result, Supplier<Statement> statementSupplier) {
-        super(result::unsafeSubscribe);
-        this.result = result;
-        this.statementSupplier = statementSupplier;
+        this(result, statementSupplier, null);
     }
 
     private DaoObservable(Observable<T> result, Supplier<Statement> statementSupplier, Action0 onTransactionCompleted) {
--- a/dao/src/main/java/se/fortnox/reactivewizard/db/transactions/DaoObservable.java
+++ b/dao/src/main/java/se/fortnox/reactivewizard/db/transactions/DaoObservable.java
@@ -3,6 +3,7 @@ package se.fortnox.reactivewizard.db.transactions;
 import rx.Observable;
 import rx.functions.Action0;
 import rx.functions.Action1;
+import se.fortnox.reactivewizard.db.statement.Statement;
 
 import java.util.concurrent.atomic.AtomicReference;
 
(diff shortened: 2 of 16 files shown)


dao/src/main/java/se/fortnox/reactivewizard/db/ObservableStatementFactory.java
```diff
@@ -59,15 +59,15 @@ private static void closeSilently(Connection connection) {
 
     public Observable<Object> create(Object[] args, ConnectionProvider connectionProvider) {
         AtomicReference<TransactionStatement> transactionHolder = new AtomicReference<>();
+        Statement dbStatement = statementFactory.create(args);
```
JonasHallFnx, line 62: I guess this is important - but I don't really understand why you need to create the statement here. This code will be running on the calling thread from the users code. Moving the creation here might change how side-effects on the `args` param will affect the running query.

The main point of this PR is to avoid running the inner Observable here? Then I guess the whole block with if `(transactionStatement != null) {...}` would not be needed any more? Then transactionHolder is not needed as well?

dao/src/test/java/se/fortnox/reactivewizard/db/DaoTransactionsTest.java
```diff
@@ -66,8 +66,12 @@ public void shouldRunOnTransactionCompletedCallback() throws SQLException {
 
         final boolean[] cbExecuted   = {false};
         DaoObservable   daoObsWithCb = ((DaoObservable)dao.updateSuccess()).doOnTransactionCompleted(() -> cbExecuted[0] = true);
+        // doOnTransactionCompleted will only be called when using createTransaction
```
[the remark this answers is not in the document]
mickej: It's also triggered every second. You are right, it should be flushed in some way for executeTransaction as well, once for every DaoObservable feels a bit overkill tho :)

dao/src/main/java/se/fortnox/reactivewizard/db/transactions/DaoObservable.java
```diff
@@ -3,38 +3,51 @@
...
 import rx.functions.Action1;
+import se.fortnox.reactivewizard.db.statement.Statement;
 
 import java.util.concurrent.atomic.AtomicReference;
 
 public class DaoObservable<T> extends Observable<T> {
 
     private final AtomicReference<TransactionStatement> transactionStatementRef;
     private final Observable<T>                         result;
+    private final Statement                             statement;
```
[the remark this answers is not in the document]
mickej: Maybe, I guess it depends on how your next comment will be solved. "Moving the creation here might change how side-effects on the args param will affect the running query."

dao/src/main/java/se/fortnox/reactivewizard/db/ObservableStatementFactory.java
mickej: Exactly, the main point is to get rid of the inner observable. But I had to keep it in order to still support DaoTransactions.createTransaction.

dao/src/main/java/se/fortnox/reactivewizard/db/transactions/TransactionStatement.java
```diff
@@ -16,6 +15,11 @@ public TransactionStatement(Transaction transaction) {
         this.statement = new AtomicReference<>();
     }
 
+    public TransactionStatement(Transaction transaction, Statement statement) {
```
rassmate, line 18: Cannot find any usage of this constructor

rassmate (changes requested):
Good job!!

But I think we can drop the createTransaction support in order to have easier code and not keep code just for backwards compatibility for these methods

rassmate (changes requested):
I think we should drop the createTransaction support in favor of easier code and to not have to maintain both ways

dao/src/main/java/se/fortnox/reactivewizard/db/transactions/Transaction.java
```diff
@@ -14,26 +14,29 @@
...
         this.daoCalls = daoCalls;
+        this.connectionProvider = connectionProvider;
     }
 
-    void add(DaoObservable daoObservable) {
+    TransactionStatement add(DaoObservable daoObservable) {
         TransactionStatement transactionStatement = new TransactionStatement(this);
         daoObservable.setTransactionStatement(transactionStatement);
         statementsToExecute.add(transactionStatement);
         statementsToSubscribe.add(transactionStatement);
```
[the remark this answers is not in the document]
mickej: Still needed in this case, see unit test _shouldRunTwoQueriesInTransactionOrderAndNotInSubscribeOrder_

```
        daoTransactions.createTransaction(find1, find2);

        find2.subscribe();
        find1.toBlocking().singleOrDefault(null);
```

dao/src/main/java/se/fortnox/reactivewizard/db/transactions/TransactionStatement.java
mickej: Oops, I was thinking about removing the add and markStatementSubscribed just as you mentioned in other comments.

dao/src/main/java/se/fortnox/reactivewizard/db/transactions/DaoTransactions.java
```diff
@@ -9,8 +9,9 @@
      * this method, when all of the Observables has been subscribed.
      *
      * @param daoCalls dao calls to run as on single transaction
-     *
+     * @deprecated Use executeTransaction instead.
      */
+    @Deprecated
     <T> void createTransaction(Observable<T>... daoCalls);
```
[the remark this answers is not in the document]
mickej: Yeah, if that can be done a lot could probably be changed with a more stream lined solution.

dao/src/main/java/se/fortnox/reactivewizard/db/transactions/Transaction.java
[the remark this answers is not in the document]
rassmate: But if you remove the create transaction this test will also be obsolete cause you cannot execute the transaction with subscribing to individual observables.

dao/src/test/java/se/fortnox/reactivewizard/db/DaoTransactionsTest.java
```diff
@@ -33,21 +34,14 @@
...
     @Test
     public void shouldRunTwoQueriesInOneTransaction() throws SQLException {
         Observable<String> find1 = dao.find();
         Observable<String> find2 = dao.find();
 
-        daoTransactions.createTransaction(find1, find2);
-
-        find1.subscribe();
-
-        db.verifyConnectionsUsed(0);
```
JonasHallFnx, line 47: You can probably keep this, to prove that calling executeTransaction does not result in connections being used, but that the `.toBlocking().singleOrDefault(null);` below will.

dao/src/test/java/se/fortnox/reactivewizard/db/DaoTransactionsTest.java
```diff
@@ -60,78 +54,84 @@ public void shouldRunTwoQueriesInOneTransaction() throws SQLException {
     }
 
     @Test
-    public void shouldRunOnTransactionCompletedCallback() throws SQLException {
-        when(db.getPreparedStatement().executeBatch())
-            .thenReturn(new int[]{1, 1});
-
-        final boolean[] cbExecuted   = {false};
-        DaoObservable   daoObsWithCb = ((DaoObservable)dao.updateSuccess()).doOnTransactionCompleted(() -> cbExecuted[0] = true);
```
JonasHallFnx, line 68: I think we need to keep this, but perhaps you could store the `Action0` inside `DaoObservable` and then just run those after the transaction has succeeded? Then you do not need to subscribe any Observables to get this job done.

dao/src/test/java/se/fortnox/reactivewizard/db/DaoTransactionsTest.java
```diff
@@ -60,78 +54,84 @@ public void shouldRunTwoQueriesInOneTransaction() throws SQLException {
...
-        inOrder.verify(connection).prepareStatement("select * from test2");
+        db.verifyConnectionsUsed(2);
+        verify(db.getConnection(), times(1)).setAutoCommit(false);
+        verify(db.getConnection(), times(1)).commit();
+        verify(db.getConnection(), times(3)).prepareStatement("select * from test");
+        verify(db.getConnection(), timeout(500).times(2)).setAutoCommit(true);
+        verify(db.getConnection(), times(2)).close();
+        verify(db.getPreparedStatement(), times(3)).close();
+
+        // TODO: Is this expected in a transaction?
```
JonasHallFnx, line 77: Well, since transactions are returning Void then there is no point in running selects in them... But sure, it should probably execute these queries anyway. Options: Keep assert / Drop assert / Change from selects to inserts and assert some other relevant call

dao/src/test/java/se/fortnox/reactivewizard/db/DaoTransactionsTest.java
```diff
@@ -201,28 +201,6 @@ public void shouldBatchWhenSameQueryInSequence() throws SQLException {
         verify(conn).close();
     }
 
-    @Test
-    public void shouldFailIfQueryIsSubscribedTwice() throws SQLException {
```
JonasHallFnx, line 205: So, then `shouldNotFailIfQueryIsSubscribedTwice`?

dao/src/test/java/se/fortnox/reactivewizard/db/DaoTransactionsTest.java
```diff
@@ -240,7 +217,18 @@ public void shouldBeAbleToUseRetryOnTransaction() throws SQLException {
 
     @Test(expected = RuntimeException.class)
     public void shouldThrowExceptionIfObservableIsNotFromDao() {
-        daoTransactions.createTransaction(Observable.empty());
+        daoTransactions.executeTransaction(Observable.empty());
+    }
+
+    @Test
+    public void shouldAllowEmptyAndNull() {
```
JonasHallFnx, line 224: Should it allow null elements in the Iterable?

dao/src/main/java/se/fortnox/reactivewizard/db/transactions/TransactionStatement.java
```diff
@@ -1,35 +1,19 @@
 package se.fortnox.reactivewizard.db.transactions;
 
-import se.fortnox.reactivewizard.db.ConnectionProvider;
 import se.fortnox.reactivewizard.db.statement.Statement;
 
 import java.sql.Connection;
 import java.sql.SQLException;
-import java.util.concurrent.atomic.AtomicReference;
 
 public class TransactionStatement implements Batchable {
```
JonasHallFnx, line 8: This is really optional, but it is possible to drop this class if you let `Statement` implement `Batchable`.

dao/src/main/java/se/fortnox/reactivewizard/db/transactions/Transaction.java
```diff
@@ -2,68 +2,41 @@
...
 public class Transaction<T> {
 
     private static final Logger                                      log                 = LoggerFactory.getLogger(Transaction.class);
-    private final        AtomicReference<ConnectionProvider>         connectionProvider  = new AtomicReference<>();
-    private final        Iterable<Observable<T>>                     daoCalls;
-    private final        ConcurrentLinkedQueue<TransactionStatement> statementsToExecute = new ConcurrentLinkedQueue<>();
-    private final        Set<TransactionStatement>                   statementsToSubscribe = new HashSet<>();
+    private final        ConnectionProvider                          connectionProvider;
+    private final        ConcurrentLinkedQueue<TransactionStatement> statementsToExecute;
     private final        AtomicBoolean                               waitingForExecution = new AtomicBoolean(true);
```
JonasHallFnx, line 19: I think you can drop this since we do no longer support asynchronous subscription of the DaoObservables.

dao/src/main/java/se/fortnox/reactivewizard/db/transactions/Transaction.java
```diff
@@ -2,68 +2,41 @@
...
-        }
-
-        Connection connection = connectionProvider.get().get();
+        Connection connection = connectionProvider.get();
         try {
             executeTransaction(connection);
             closeConnection(connection);
-            allCompleted();
-        } catch (Throwable e) {
+        } catch (Exception e) {
```
JonasHallFnx, line 35: I think this needs to be Throwable to avoid that a non-exception error fails to rollback the transaction?

dao/src/main/java/se/fortnox/reactivewizard/db/transactions/Transaction.java
```diff
@@ -2,68 +2,41 @@
...
-            // all DaoObservables have not been subscribed yet
-            throw new RuntimeException("Transaction execute called before all Observables were subscribed. This should never happen.");
-        }
-
-        if (isModifiedAfterCreation()) {
-            throw new RuntimeException("Transaction cannot be modified after creation.");
-        }
-
+    public void execute() throws Exception {
         if (!waitingForExecution.compareAndSet(true, false)) {
```
JonasHallFnx, line 27: drop this?

dao/src/test/java/se/fortnox/reactivewizard/db/DaoTransactionsTest.java
```diff
@@ -33,21 +36,18 @@
...
-        daoTransactions.createTransaction(find1, find2);
-
-        find1.subscribe();
-
+        Observable<Void> executeTransactionObs = daoTransactions.executeTransaction(find1, find2);
         db.verifyConnectionsUsed(0);
         verify(db.getConnection(), times(0)).prepareStatement(any());
 
-        find2.toBlocking().singleOrDefault(null);
+        executeTransactionObs.toBlocking().subscribe();
```
rassmate, line 50: A small possibility of a race condition here since you start verifying stuff before you are certain it has executed.
Perhaps use toBlocking() to be 100% sure

dao/src/main/java/se/fortnox/reactivewizard/db/transactions/DaoObservable.java
```diff
@@ -3,38 +3,52 @@
...
 import rx.functions.Action1;
+import se.fortnox.reactivewizard.db.statement.Statement;
 
-import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Supplier;
 
 public class DaoObservable<T> extends Observable<T> {
+    private final Observable<T> result;
+    private final Supplier<Statement> statementSupplier;
+    private Action0 onTransactionCompleted;
```
JonasHallFnx, line 13: Can be final?

JonasHallFnx (approved):
Great job in this PR! You removed alot of old complicated stuff. Kudos for keeping the tests relatively intact, it gives me confidence in that your changes are working as expected.

rassmate (approved):

rassmate:
🌟

PR #263, opened 2021-11-29
logging onErrorDropped at a different loglevel.

rassmate:
Applying this will cause onErrorDropped to be logged with a lower log level.

onErrorDropped is called whenever an onError signal is sent to a subscriber but that subscriber has already gotten an exception earlier.

```java
Mono.fromCallable(() -> {
                Thread.sleep(60);
                return "Success";
            })
            .subscribeOn(Schedulers.elastic())
            .timeout(Duration.ofMillis(50))
            .onErrorResume(throwable -> Mono.just("Timeout"))
            .doOnSuccess(result -> System.out.println("Result: " + result))
            .block();
```


tsfcode (approved):

JonasHallFnx (approved):

bootstrap/src/main/java/se/fortnox/reactivewizard/SetupModule.java
```diff
@@ -0,0 +1,33 @@
...
+
+    @Override
+    public void configure(Binder binder) {
+        Hooks.onErrorDropped(throwable -> {
+            logError(LOG, throwable);
+        });
+    }
+
+    /**
+     * package private to be able to be tested without too much trouble.
```
jepp3, line 23: maybe use ` com.google.common.annotations.VisibleForTesting;`  if  linter allows 

```java
@VisibleForTesting
void logError(Logger logger, Throwable throwable) {
```

bootstrap/src/main/java/se/fortnox/reactivewizard/SetupModule.java
```diff
@@ -0,0 +1,33 @@
...
+     *
+     * @param logger the logger to be used
+     * @param throwable the exception
+     */
+    void logError(Logger logger, Throwable throwable) {
+        logger.warn("Tried to send item or error to subscriber but the subscriber had already left. " +
```
jepp3, line 29: unsure if this would be cleaner: 


The error has been dropped because it cannot be propagated through the reactive chain.  This happens when a publisher calls onError on the operator despite having already called onCompleted on it previously. This may happen when an error is generated by different places in the reactive chain and only one of them can be propagated and resulting in this exception being dropped.

jepp3 (approved):

bootstrap/src/main/java/se/fortnox/reactivewizard/SetupModule.java
rassmate: I'll merge our two messages

PR #266, opened 2021-12-03
Fix support for hot reload of dao-method annotations

JonasHallFnx:
When you recompile a class during a debug session, it will be hot reloaded and replace the existing class. However, dao instances that has already been injected will still be using the old version of the interface and queries will therefore not update. This PR fixes that bug and allows for hot reloading to work again.

### DIFF
--- a/utils/src/test/java/se/fortnox/reactivewizard/util/ReflectionUtilTest.java
+++ b/utils/src/test/java/se/fortnox/reactivewizard/util/ReflectionUtilTest.java
@@ -299,6 +299,12 @@ public class ReflectionUtilTest {
         assertThat(method).isEqualTo(TestResource.class.getMethod("resourceGet", String.class, String.class));
     }
 
+    @Test
+    public void shouldGetRedefinedVersionOfMethod() throws NoSuchMethodException {
+        Method method = TestResource.class.getMethod("resourceGet", String.class, String.class);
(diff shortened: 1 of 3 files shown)


utils/src/test/java/se/fortnox/reactivewizard/util/ReflectionUtilTest.java
```diff
@@ -299,6 +299,12 @@ public void shouldSupportCglibProxiesWhenLocatingOverriddenMethods() throws Exce
         assertThat(method).isEqualTo(TestResource.class.getMethod("resourceGet", String.class, String.class));
     }
 
+    @Test
+    public void shouldGetRedefinedVersionOfMethod() throws NoSuchMethodException {
```
JonasHallFnx, line 303: This just test that you will get the same method. Testing a new version of the exact same class is beyond my skills. You would have to use CGLib or ASM to pull that off, and we even have code inside `getUserDefinedClass` which circumvents that, so I have no clue how to write a test that shows that hot reload is really working.

Tested it manually though.

flowertwig (approved):

mickej (approved):

j-holmstrom (approved):

tsfcode (approved):

rassmate (approved):

rassmate (approved):

PR #269, opened 2021-12-23
WIP - Feature/log4j2 from config files

j-holmstrom:
Migrated to log4j2. We adapted log42j configuration instead of having it in the config.yml with loggingfactory. The reason for this was that we want to do the logging configuration more flexible without requiring changes to the framework. As well we could remove a lot of code that just converted between our format and the format log4j.


jepp3 (approved):

rassmate (approved):

JonasHallFnx (approved):

rassmate (changes requested):
Removing the logging module also removes the central place for specifying which version of slf4j we are using.

So wherever we remove the logging-dependency, we should add a dependency to slf4j-api module so the code does not get the dependency from some shady lib we didnt expect.

Perhaps add it to the dependencyManagement section of the root-pom to avoid version conflicts.

rassmate (approved):

PR #274, opened 2022-01-10
Adding a logging verifier junit rule to avoid being forced to set the…

rassmate:
… appender up and down.


rassmate:
Just a spontaneous idea.

PR #276, opened 2022-01-14
Using first available ConnectionScheduler where connectionProvider is set.

rassmate:
having a ConnectionScheduler with a connectionProvider set.
A daoobservable could have connectionProvider null when created from a DbProxy without a connectionProvier set.

### DIFF
--- a/dao/src/main/java/se/fortnox/reactivewizard/db/transactions/ConnectionScheduler.java
+++ b/dao/src/main/java/se/fortnox/reactivewizard/db/transactions/ConnectionScheduler.java
@@ -15,6 +15,10 @@ public class ConnectionScheduler {
         this.scheduler = scheduler;
     }
 
+    boolean hasConnectionProvider() {
+        return connectionProvider != null;
+    }
+
     public void schedule(Subscriber<?> subscription, ThrowableAction action) {
         Scheduler.Worker worker = scheduler.createWorker();
         worker.schedule(() -> {
--- a/dao/src/main/java/se/fortnox/reactivewizard/db/transactions/DaoTransactionsImpl.java
+++ b/dao/src/main/java/se/fortnox/reactivewizard/db/transactions/DaoTransactionsImpl.java
@@ -68,8 +68,9 @@ public class DaoTransactionsImpl implements DaoTransactions {
 
     private <T> ConnectionScheduler getConnectionScheduler(Collection<Observable<T>> daoCallsCopy) {
         Observable<T> daoObservable = daoCallsCopy.stream()
+            .filter(obs -> ((DaoObservable<T>)obs).getStatementContextSupplier().get().getConnectionScheduler().hasConnectionProvider())
             .findFirst()
-            .orElseThrow(() -> new RuntimeException("No DaoObservable found"));
+            .orElseThrow(() -> new RuntimeException("No DaoObservable with a valid connection provider was found"));
 
         return ((DaoObservable<T>) daoObservable).getStatementContextSupplier().get().getConnectionScheduler();
     }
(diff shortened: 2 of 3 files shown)


dao/src/main/java/se/fortnox/reactivewizard/db/transactions/DaoTransactionsImpl.java
```diff
@@ -68,8 +68,9 @@ public DaoTransactionsImpl() {}
 
     private <T> ConnectionScheduler getConnectionScheduler(Collection<Observable<T>> daoCallsCopy) {
         Observable<T> daoObservable = daoCallsCopy.stream()
+            .filter(obs -> ((DaoObservable<T>)obs).getStatementContextSupplier().get().getConnectionScheduler().hasConnectionProvider())
```
JonasHallFnx, line 71: If you want to, you could do a map and then filter, avoiding duplicated `.getStatementContextSupplier().get().getConnectionScheduler()`

dao/src/main/java/se/fortnox/reactivewizard/db/transactions/ConnectionScheduler.java
```diff
@@ -15,6 +15,10 @@ public ConnectionScheduler(ConnectionProvider connectionProvider, Scheduler sche
         this.scheduler = scheduler;
     }
 
+    boolean hasConnectionProvider() {
```
JonasHallFnx, line 18: Since getStatementContextSupplier() returns an Optional, would it be possible to use this fact to avoid setting that statementContextProvider to a value, for the case when there is no connection provider?

dao/src/test/java/se/fortnox/reactivewizard/db/DaoTransactionsTest.java
```diff
@@ -367,6 +370,35 @@ public void shoulUseSpecifiedConnectionAndScheduler() throws SQLException {
...
+            null, new DbStatementFactoryFactory(), new JsonSerializerFactory());
+        DbProxy dbProxyWithConnectionProvider = dbProxyWithoutConnectionProvider.usingConnectionProvider(otherConnectionProvider);
+
+        when(secondDb.getPreparedStatement().executeBatch())
+            .thenReturn(new int[]{1, 1});
+
+        TestDao daoWithConnectionProvider = dbProxyWithConnectionProvider.create(TestDao.class);
+        TestDao daoWithoutConnectionProvider = dbProxyWithoutConnectionProvider.create(TestDao.class);
+
+        daoTransactions.executeTransaction(daoWithoutConnectionProvider.updateSuccess(), daoWithConnectionProvider.updateSuccess())
```
JonasHallFnx, line 390: Successful transaction needs to be verified here. Since it is not blocking subscribe, it is not implicit.

dao/src/test/java/se/fortnox/reactivewizard/db/DaoTransactionsTest.java
```diff
@@ -367,6 +370,35 @@ public void shoulUseSpecifiedConnectionAndScheduler() throws SQLException {
...
+            .subscribeOn(secondScheduler).subscribe();
+        secondScheduler.triggerActions();
+
+        secondDb.verifyConnectionsUsed(1);
+        db.verifyConnectionsUsed(0);
+
+        assertThatExceptionOfType(RuntimeException.class)
+            .isThrownBy(() -> daoTransactions.executeTransaction(daoWithoutConnectionProvider.updateSuccess()).subscribe())
```
JonasHallFnx, line 398: This subscribe should be asynchronous (not blocking) so an exception should be sent to subscriber rather than thrown to caller?

dao/src/main/java/se/fortnox/reactivewizard/db/transactions/ConnectionScheduler.java
rassmate: It is not an optional, it is a supplier. And inside there is the Statement we want to run so I am not so sure about this.

JonasHallFnx (approved):

flowertwig (approved):

PR #289, opened 2022-01-28
Added possibility to mask headers in logging

flowertwig:


jaxrs/src/main/java/se/fortnox/reactivewizard/jaxrs/RequestLogger.java
```diff
@@ -61,24 +69,33 @@ public static void headersToString(HttpServerResponse response, StringBuilder lo
      * @param header The header to get the value from
      * @return value of the header or "REDACTED" if the header contains sensitive information
      */
-    public static String getHeaderValueOrRedact(Map.Entry<String, String> header) {
-        return getHeaderValueOrRedact(header, emptySet());
+    public String getHeaderValueOrRedactIncoming(Map.Entry<String, String> header) {
```
JonasHallFnx, line 72: Incoming/Outgoing might be problematic because both client and server has both a request and a response with headers. Is this intentional (outgoing being response of server and request of client) or not? Otherwise I suggest using "server"/"client".

client/src/test/java/se/fortnox/reactivewizard/client/HttpClientTest.java
```diff
@@ -689,7 +698,7 @@ public void shouldHandleLargeResponses() {
     public void shouldLogErrorOnTooLargeResponse() {
         DisposableServer server = startServer(HttpResponseStatus.OK, generateLargeString(11));
         try {
-            Appender                           mockAppender = LoggingMockUtil.createMockedLogAppender(HttpClient.class);
+            Appender                           mockAppender = createMockedLogAppender(HttpClient.class);
```
rassmate, line 701: LoggingVerifier?

client/src/test/java/se/fortnox/reactivewizard/client/HttpClientTest.java
```diff
@@ -708,15 +717,16 @@ public void shouldLogErrorOnTooLargeResponse() {
...
         }
     }
 
     @Test
     public void shouldRedactAuthorizationHeaderInLogsAndExceptionMessage() throws Exception {
         DisposableServer server = startServer(BAD_REQUEST, "someError");
+
         try {
-            Appender                           mockAppender = LoggingMockUtil.createMockedLogAppender(HttpClient.class);
+            Appender                           mockAppender = createMockedLogAppender(HttpClient.class);
```
rassmate, line 729: LoggingVerifier?

rassmate (approved):

jaxrs/src/main/java/se/fortnox/reactivewizard/jaxrs/RequestLogger.java
```diff
@@ -141,22 +157,57 @@ public static void logAccess(HttpServerRequest request, HttpServerResponse respo
...
             log.info(logLine.toString());
         }
     }
 
-    public void log(HttpServerRequest request, HttpServerResponse response, long requestStartTime) {
+    public void log(Logger logger, HttpServerRequest request, HttpServerResponse response, long requestStartTime) {
         logRequestResponse(request, response, requestStartTime, logger);
     }
+
+    public void addHeaderTransformationOutgoing(String header, UnaryOperator<String> transformation) {
```
jepp3, line 179: when configuring the transformation of headers in the application, are we supposed to use this method and the one for incoming?
If so, should we not document their purpose for devs of the framework?

jaxrs/src/main/java/se/fortnox/reactivewizard/jaxrs/RequestLogger.java
```diff
@@ -1,28 +1,36 @@
...
-import java.util.stream.Collectors;
+import java.util.TreeMap;
+import java.util.function.UnaryOperator;
 
-import static java.util.Collections.emptySet;
+import static java.util.Objects.requireNonNull;
+import static java.util.stream.Collectors.toMap;
 
 /**
  * Logs incoming requests to a Logger.
```
jepp3, line 21: Should we rewrite the term _incoming_ here as well? and is the request logger only used for requests handled by the server and not the client? :thinking:

jaxrs/src/main/java/se/fortnox/reactivewizard/jaxrs/RequestLogger.java
flowertwig: server/client is better.

jaxrs/src/main/java/se/fortnox/reactivewizard/jaxrs/RequestLogger.java
flowertwig: Yes, we really should. Odd that checkstyle didn't complain as it did on so much else. :P

jepp3 (approved):

PR #292, opened 2022-02-24
Add possibility to turn off lambda generation

JonasHallFnx:
By starting the system with -DuseLambdas=false all reflection will use standard reflection calls instead of compiling lambdas.

### DIFF
--- a/utils/src/main/java/se/fortnox/reactivewizard/util/LambdaCompiler.java
+++ b/utils/src/main/java/se/fortnox/reactivewizard/util/LambdaCompiler.java
@@ -1,0 +1,77 @@
+package se.fortnox.reactivewizard.util;
+
+import java.lang.invoke.CallSite;
+import java.lang.invoke.LambdaMetafactory;
+import java.lang.invoke.MethodHandle;
+import java.lang.invoke.MethodHandles;
+import java.lang.invoke.MethodType;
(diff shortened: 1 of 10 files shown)


utils/src/main/java/se/fortnox/reactivewizard/util/LambdaCompiler.java
```diff
@@ -0,0 +1,77 @@
...
+
+public class LambdaCompiler {
+    static boolean useLambdas = "true".equals(System.getProperty("useLambdas", "true"));
+
+    static <T> Supplier<T> compileLambdaSupplier(MethodHandles.Lookup lookup, MethodHandle methodHandle) throws Throwable {
+        if (!useLambdas) {
+            return () -> {
+                try {
+                    return (T) methodHandle.invoke();
+                } catch (Throwable e) {
```
jepp3, line 20: why catch on throwable? 
Error and its sub-classes are always unchecked exceptions, while the sub-classes of Exception can be either checked or unchecked. If the idea is to make checked exceptions unchecked and not to catch OOMs and other erros, then I guess you don't need throwable here.

jepp3 (approved):

flowertwig (approved):

rassmate (approved):

PR #295, opened 2022-03-07
Use static logger in main

rassmate:
Use static logger in the Main class instead of fetching on demand.

Fetching the logger on demand at exception caused no output to be shown from the exception


flowertwig (approved):

jepp3 (approved):
:+1:

PR #300, opened 2022-04-06
To allow streaming results with CollectionOptions

rassmate:
we need the CollectionOptions isLastRecord to be default false.

To allow this we need to change the strategy of only setting it to false when we have more items to always set the correct value onComplete in PagingOperator

### DIFF
--- a/api/src/main/java/se/fortnox/reactivewizard/CollectionOptions.java
+++ b/api/src/main/java/se/fortnox/reactivewizard/CollectionOptions.java
@@ -20,7 +20,7 @@ public class CollectionOptions {
     @QueryParam("order")
     private SortOrder order;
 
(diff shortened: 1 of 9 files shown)


api/src/main/java/se/fortnox/reactivewizard/CollectionOptions.java
```diff
@@ -20,7 +20,7 @@
     @QueryParam("order")
     private SortOrder order;
 
-    private boolean lastRecord = true;
+    private boolean lastRecord = false;
```
JonasHallFnx, line 23: Is it possible to test the desired behaviour? A test with a streaming result and two pages should do?

JonasHallFnx (approved):

flowertwig (approved):

PR #323, opened 2022-05-05
Ignore non-standard versions of jackson-bom

flowertwig:
Also added explicit reviewers since team doesn't seem to work.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -11,3 +11,10 @@ updates:
       interval: "weekly"
     reviewers:
       - "FortnoxAB/PlatMap"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -11,3 +11,10 @@ updates:
       interval: "weekly"
     reviewers:
       - "FortnoxAB/PlatMap"
```
jepp3, line 13: maybe we should use the same casing used when tagging the group in a conversation` @FortnoxAB/platmap` @FortnoxAB/platmap

jepp3 (approved):

PR #355, opened 2022-07-05
Add type info on deserialization errors

ivarfnox:
Improve troubleshooting bad type serialization in framework. These errors might occur for example when migrating from old `jackson-annotations` versions that move from the package `javax.validation` to `jakarta.validation`.

### DIFF
--- a/client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
+++ b/client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
@@ -590,9 +590,14 @@ public class HttpClient implements InvocationHandler {
 
         try {
             JavaType     javaType = TypeFactory.defaultInstance().constructType(type);
-            ObjectReader reader   = objectMapper.readerFor(javaType);
-            Object       value    = reader.readValue(string);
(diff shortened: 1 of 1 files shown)


rassmate (approved):

client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
```diff
@@ -590,9 +590,14 @@ protected Mono<Object> deserialize(Method method, String string) {
...
         try {
             JavaType     javaType = TypeFactory.defaultInstance().constructType(type);
-            ObjectReader reader   = objectMapper.readerFor(javaType);
-            Object       value    = reader.readValue(string);
-            return Mono.justOrEmpty(value);
+            try {
+                ObjectReader reader = objectMapper.readerFor(javaType);
+                Object       value    = reader.readValue(string);
+                return Mono.justOrEmpty(value);
+            } catch (NoClassDefFoundError e) {
```
rassmate, line 597: A bit weird to catch this error. But i don't see any harm.

rassmate (approved):

PR #375, opened 2022-08-17
Support for JUnit 5

softqwewasd:


test/pom.xml
```diff
@@ -17,6 +17,11 @@
             <artifactId>rxjava</artifactId>
         </dependency>
 
+        <dependency>
```
flowertwig, line 20: We should remove this if not needed.

flowertwig (commented):
I think we should migrate one or two test classes to junit5 just to show that it works.

rassmate (approved):
GJ!

test/src/test/java/se/fortnox/reactivewizard/examples/junit5_examples/Junit5Tests.java
```diff
@@ -0,0 +1,35 @@
...
+
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.CsvSource;
+
+import java.util.function.Supplier;
+
+class Junit5Tests {
```
jepp3, line 11: I think we should remove these pure example test cases. We can state in the reactive wizard documentation in backstage that junit5 is supported out of the box maybe.

jepp3 (changes requested):
I think we should remove these pure example test cases. We can state in the reactive wizard documentation in backstage that junit5 is supported out of the box maybe.

softqwewasd:
The example code has been removed. I migrated a test class in the _jaxrs_ module from JUnit 4 to JUnit 5 since I could not find tests in the _test_ module that are fit to be migrated into an example parameterized test.

jepp3 (approved):

flowertwig (approved):

test/src/test/java/se/fortnox/reactivewizard/test/LoggingInfoVerifierTest.java
```diff
@@ -3,6 +3,7 @@
 import org.apache.logging.log4j.Level;
 import org.junit.Rule;
 import org.junit.Test;
+import org.junit.jupiter.params.ParameterizedTest;
```
rassmate, line 6: unused import?

rassmate (approved):

PR #376, opened 2022-08-17
Added an error message when trying to return null values from the database as part of the reactive stream.

jepp3:
The reactive streams specification disallows null values in a sequence. Now when the dao module has been migrated to project reactor we cannot return null values in a reactive stream. We do that when we select specific columns from a database, that may be null.  Consider the following:

```
@Select( "select may_be_null_column from table where id = :id" )
Flux<String> getStuff(Integer id);
```
Here may_be_null_column may be null.

Backward compatibility is hard to get here.  we can either: 
* Ignore the null values returned in the sequence
* Throw an exception stating that returning null values cannot be done. 

We went with the "throw an exception" since ignoring values might be more dangerous because it will hide the problem. As well the user can get around the issue by wrapping the possible null values. 

So one can say the dao migration is breaking because it will changed the behavior on null values. This will only make it clearer for developers to understand the problem.

### DIFF
--- a/dao/src/main/java/se/fortnox/reactivewizard/db/statement/SelectStatementFactory.java
+++ b/dao/src/main/java/se/fortnox/reactivewizard/db/statement/SelectStatementFactory.java
@@ -28,8 +28,8 @@ public class SelectStatementFactory extends AbstractDbStatementFactory {
             try (ResultSet resultSet = statement.executeQuery()) {
                 while (resultSet.next()) {
                     if (fluxSink != null) {
-                        var val = deserializer.deserialize(resultSet);
-                        if(val == null){
+                        var value = deserializer.deserialize(resultSet);
+                        if (value == null) {
                             throw new NullPointerException("""
                                 One or more selected values from the database is null.
                                 Project Reactor does not allow emitting null values in a stream. Wrap the return value from the dao interface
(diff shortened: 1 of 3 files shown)


flowertwig (approved):

rassmate (approved):

dao/src/main/java/se/fortnox/reactivewizard/db/statement/SelectStatementFactory.java
```diff
@@ -25,7 +28,19 @@ protected void executeStatement(Connection connection, Object[] args, FluxSink f
             try (ResultSet resultSet = statement.executeQuery()) {
                 while (resultSet.next()) {
                     if (fluxSink != null) {
-                        fluxSink.next(deserializer.deserialize(resultSet));
+                        var value = deserializer.deserialize(resultSet);
+                        if (value == null) {
+                            throw new NullPointerException("""
```
JonasHallFnx, line 33: Is it possible to add the query and/or typeinfo of the method to the error, otherwise you have no clue to what query is causing this?

jepp3: good point, I added the parameterized query.

JonasHallFnx (approved):

PR #377, opened 2022-08-19
Return 400 HTTP response code when the query contains an invalid hex …

softqwewasd:


jaxrs/src/main/java/se/fortnox/reactivewizard/jaxrs/JaxRsRequest.java
```diff
@@ -108,8 +108,13 @@ public String getQueryParam(String key) {
     public String getQueryParam(String key, String defaultValue) {
         if (queryParameters == null) {
             QueryStringDecoder queryStringDecoder = new QueryStringDecoder(req.uri());
-            queryParameters = queryStringDecoder.parameters();
+            try {
+                queryParameters = queryStringDecoder.parameters();
+            } catch (IllegalArgumentException e) {
+                throw new WebException(HttpResponseStatus.BAD_REQUEST, "invalidhexbyte", "Failed to decode query params due to invalid two digit hexadecimal");
```
jepp3, line 114: How do we know that the IllegalArgumentException is caused by what we state and assume here?

softqwewasd: I made the assumption that IllegalArgumentException is thrown if and only if the hexbyte decoding fails. I could not find other places where an IllegalArgumentException would be thrown for some query param values other than the hexbyte decoding part of the code. That said, I do not think we should pass the exception message to WebException but we can change the message to something more general, like "Failed to decode query parameters" and the "invalidhexbyte" can be changed to "querydecodefailure".

flowertwig: Will any of that (except the status code) be exposed in the http response? If so, I don't think we should include it in the WebException, but log the IllegalArgumentException instead.

flowertwig (changes requested):

jaxrs/src/main/java/se/fortnox/reactivewizard/jaxrs/JaxRsRequest.java
```diff
@@ -108,8 +108,14 @@ public String getQueryParam(String key) {
     public String getQueryParam(String key, String defaultValue) {
         if (queryParameters == null) {
             QueryStringDecoder queryStringDecoder = new QueryStringDecoder(req.uri());
-            queryParameters = queryStringDecoder.parameters();
+            try {
+                queryParameters = queryStringDecoder.parameters();
+            } catch (IllegalArgumentException e) {
+                LOG.error(e.getMessage());
```
flowertwig, line 114: We should have a log message that briefly explains what happened and then include the exception as the last param in LOG.error. Thus we will get the message and stacktrace.

flowertwig (changes requested):

jaxrs/src/main/java/se/fortnox/reactivewizard/jaxrs/JaxRsRequest.java
```diff
@@ -108,8 +108,14 @@ public String getQueryParam(String key) {
     public String getQueryParam(String key, String defaultValue) {
         if (queryParameters == null) {
             QueryStringDecoder queryStringDecoder = new QueryStringDecoder(req.uri());
-            queryParameters = queryStringDecoder.parameters();
+            try {
+                queryParameters = queryStringDecoder.parameters();
+            } catch (IllegalArgumentException e) {
+                LOG.error("Failed to decode HTTP query params for request " + req.method().name() + " " + req.uri(), e);
```
flowertwig, line 114: Use {} placeholders for the two params.

jaxrs/src/test/java/se/fortnox/reactivewizard/jaxrs/JaxRsResourceTest.java
```diff
@@ -746,6 +746,21 @@ public void shouldAcceptBeanParamInherited() {
...
+        assertBadRequest("/test/accepts?%A");
+        assertBadRequest("/test/accepts?%=");
+        assertBadRequest("/test/accepts?%?");
+        assertBadRequest("/test/accepts?%AG");
+        assertBadRequest("/test/accepts?%A@");
+    }
+
+    private void assertBadRequest(String uri) {
+        MockHttpServerResponse mockHttpServerResponse = get(service, uri);
+        assertThat(mockHttpServerResponse.status()).isEqualTo(HttpResponseStatus.BAD_REQUEST);
```
flowertwig, line 761: Use LoggingVerifier to assert the logging as well.

flowertwig (changes requested):

jaxrs/src/test/java/se/fortnox/reactivewizard/jaxrs/JaxRsResourceTest.java
softqwewasd: :+1:

flowertwig (approved):

PR #404, opened 2022-09-20
execute transaction will now accept Mono and Flux.

rassmate:
Since commands sent to a database is typically inserts or updates which normally would return a single integer (number of rows affected) or empty if specified as Mono<Void>

### DIFF
--- a/dao/src/main/java/se/fortnox/reactivewizard/db/transactions/DaoTransactionsFlux.java
+++ b/dao/src/main/java/se/fortnox/reactivewizard/db/transactions/DaoTransactionsFlux.java
@@ -1,5 +1,6 @@
 package se.fortnox.reactivewizard.db.transactions;
 
+import org.reactivestreams.Publisher;
 import reactor.core.publisher.Mono;
 
 public interface DaoTransactionsFlux {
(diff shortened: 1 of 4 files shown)


dao/src/main/java/se/fortnox/reactivewizard/db/transactions/DaoTransactionsFlux.java
```diff
@@ -1,6 +1,6 @@
 package se.fortnox.reactivewizard.db.transactions;
 
-import reactor.core.publisher.Flux;
+import reactor.core.publisher.Mono;
 
 public interface DaoTransactionsFlux {
```
flowertwig, line 5: Shouldn't we change this name? `DaoTransactionsReactor` and then replace `DaoTransactions` once rxjava support is removed?

rassmate: If we are going to change it in the future, perhaps we can leave it like this and change it to the correct one once its name is going to be DaoTransactions?

JonasHallFnx (approved):

flowertwig (approved):

utils/src/main/java/se/fortnox/reactivewizard/util/ReactiveDecorator.java
```diff
@@ -72,6 +72,24 @@ public static <T> Optional<T> getDecoration(Flux<?> wrapper) {
         return Optional.of((T) ((DecoratedFlux)wrapper).getDecoration());
     }
 
+    /**
+     * Get the decoration of a Flux.
```
flowertwig, line 76: Flux -> Publisher

PR #412, opened 2022-10-11
Feature/migrate dao module to reactor

softqwewasd:


dao/src/test/java/se/fortnox/reactivewizard/db/ParameterizedQueryTest.java
```diff
@@ -23,7 +30,7 @@ public class ParameterizedQueryTest {
 
     @Test
     public void shouldResolveParametersFromQuery() throws SQLException {
-        dao.namedParameters("myid", "myname").toBlocking().singleOrDefault(null);
+        dao.namedParameters("myid", "myname").blockFirst();
```
flowertwig, line 33: blockLast for all Flux methods?

dao/src/test/java/se/fortnox/reactivewizard/db/ParameterizedQueryTest.java
```diff
@@ -60,7 +67,7 @@ public void shouldResolveParametersWithoutAnnotationFromQuery() throws SQLExcept
     @Test
     public void shouldThrowExceptionIfUnnamedParamsUsedInQuery() {
         try {
-            dao.unnamedParameters("myid", "myname").toBlocking().singleOrDefault(null);
+            dao.unnamedParameters("myid", "myname").blockFirst();
```
flowertwig, line 70: We could use StepVerifier to assert exceptions.

dao/src/test/java/se/fortnox/reactivewizard/db/DbProxyTest.java
```diff
@@ -45,12 +49,73 @@ public class DbProxyTest {
...
+    }
+
+    @Test
+    public void shouldSignalCompleteWhenQueryReturnsNoResultOnMonoReturnType() {
+        StepVerifier.create(dbProxyTestDao.selectMultipleWithMono()).verifyComplete();
+    }
+
+    @Test
+    public void shouldSignalErrorWhenDaoMethodDoesNotReturnAPublisher() {
+        try {
```
flowertwig, line 87: use
`assertThatException()
            .describedAs("Expected RuntimeException with the message...")
            .isThrownBy(() -> )
            .withMessage(bla)`

dao/src/test/java/se/fortnox/reactivewizard/db/DbProxyTest.java
```diff
@@ -45,12 +49,73 @@ public class DbProxyTest {
...
+            loggingVerifier.verify(Level.ERROR, "DAO method " +
+                "se.fortnox.reactivewizard.db.DbProxyTestDao::selectObservable " +
+                "must return a Flux or Mono. Found rx.Observable");
+        }
+    }
+
+    @Test
+    public void shouldSignalNextAndCompleteWhenSuccessfullyClosedConnectionOnMonoReturnType() throws SQLException {
+        mockDb.addRowColumn(1, 1, "sql_val", String.class, "myname");
+        StepVerifier.create(dbProxyTestDao.select("myname")).expectNextCount(1).verifyComplete();
```
flowertwig, line 106: Matter of taste perhaps, but I think readability improves with newlines for expect... and verify...

dao/src/test/java/se/fortnox/reactivewizard/db/DbProxyTest.java
```diff
@@ -410,57 +478,48 @@ public void testAllPathsOfTryWithResourceUsingCount() throws SQLException {
...
     }
 
     private void insertAndAssertError(Flux<?> insert, String message) {
         assertThatExceptionOfType(Exception.class)
                 .isThrownBy(insert::blockLast)
                 .satisfies(exception -> {
                     Throwable selfOrCause = Optional.ofNullable(exception.getCause()).orElse(exception);
-                    assertThat(selfOrCause.getMessage()).contains(message);
+                    if (Objects.nonNull(message)) {
+                        assertThat(selfOrCause.getMessage()).contains(message);
```
flowertwig, line 499: assertj goodness: hasMessageContaining :D 
Why do we want to send null message btw?

dao/src/test/java/se/fortnox/reactivewizard/db/DbProxyTest.java
```diff
@@ -410,57 +478,48 @@ public void testAllPathsOfTryWithResourceUsingCount() throws SQLException {
...
         assertThatNoException()
-                .isThrownBy(() -> dbProxyTestDao.updateAllowingZeroAffectedRows("mykey", "myval").toBlocking().singleOrDefault(null));
+                .isThrownBy(() -> dbProxyTestDao.updateAllowingZeroAffectedRows("mykey", "myval").block());
     }
 
     @Test
     public void shouldAllowVoidReturnType() throws SQLException {
         mockDb.setUpdatedRows(1);
-        assertThat(dbProxyTestDao.insertVoidReturnType().isEmpty().toBlocking().single()).isTrue();
+        assertThat(dbProxyTestDao.insertVoidReturnType().hasElement().block()).isFalse();
```
flowertwig, line 514: Isn't this a test of the Void type? Is it possible for a Mono<Void> to have an element?

dao/src/test/java/se/fortnox/reactivewizard/db/DbProxyTest.java
```diff
@@ -410,57 +478,48 @@ public void testAllPathsOfTryWithResourceUsingCount() throws SQLException {
...
         verify(mockDb.getPreparedStatement()).executeUpdate();
     }
 
     @Test
     public void shouldFailIfMinimumAffectedRowsNotReachedForVoidReturnType() throws SQLException {
         mockDb.setUpdatedRows(0);
         try {
-            dbProxyTestDao.insertVoidReturnType().isEmpty().toBlocking().single();
+            dbProxyTestDao.insertVoidReturnType().block();
```
flowertwig, line 522: we should use assertj here

dao/src/test/java/se/fortnox/reactivewizard/db/DbProxyTest.java
```diff
@@ -470,25 +529,29 @@ public void shouldFailIfMinimumAffectedRowsNotReachedForVoidReturnType() throws
...
-
-        testSubscriber.requestMore(1);
-        Thread.sleep(100);
-        testSubscriber.assertValueCount(1);
-        testSubscriber.assertNoTerminalEvent();
 
-        verify(mockDb.getConnection()).close();
-        verify(mockDb.getPreparedStatement()).close();
-        verify(mockDb.getResultSet()).close();
+        StepVerifier.create(dbProxyTestDao.selectFlux("mykey"))
```
flowertwig, line 536: Nice!

dao/src/test/java/se/fortnox/reactivewizard/db/DbProxyTest.java
```diff
@@ -470,25 +529,29 @@ public void shouldFailIfMinimumAffectedRowsNotReachedForVoidReturnType() throws
...
-        testSubscriber.requestMore(2);
-        Thread.sleep(100);
-        testSubscriber.assertValueCount(2);
-        testSubscriber.assertCompleted();
+    private void assertMockDbClosed() {
+        try {
+            verify(mockDb.getConnection()).close();
+            verify(mockDb.getPreparedStatement()).close();
+            verify(mockDb.getResultSet()).close();
+        } catch (SQLException e) {
```
flowertwig, line 552: When does this happen?

dao/src/test/java/se/fortnox/reactivewizard/db/DbProxyTest.java
softqwewasd: There are some tests that want to skip the exception message assertion, and they do so by setting it to null, see [https://github.com/FortnoxAB/reactive-wizard/blob/master/dao/src/test/java/se/fortnox/reactivewizard/db/DbProxyTest.java#L430](DbProxyTest:430). Not sure why though.

dao/src/test/java/se/fortnox/reactivewizard/db/DbProxyTest.java
softqwewasd: This happens in _shouldSupportBackpressure_ test. I just moved the code to another method. Other than this test, I check that the connection closes in some of the new tests added in this PR.

dao/src/test/java/se/fortnox/reactivewizard/db/ParameterizedQueryTest.java
softqwewasd: The method seems to just throw the exception instead of error signaling. We can use Assertj though.

dao/src/test/java/se/fortnox/reactivewizard/db/DaoTest.java
```diff
@@ -1,64 +1,49 @@
...
 
         db.verifyConnectionsUsed(2);
         verify(db.getConnection(), times(2)).setAutoCommit(true);
         verify(db.getConnection(), never()).commit();
-        verify(db.getConnection(), times(1)).prepareStatement("select * from test");
-        verify(db.getConnection(), times(1)).prepareStatement("update foo set other_key=val");
-        verify(db.getConnection(), times(2)).close();
+        verify(db.getConnection()).prepareStatement("select * from test");
+        verify(db.getConnection()).prepareStatement("update foo set other_key=val");
+        verify(db.getConnection()).close();
```
flowertwig, line 45: What caused the change from 2 to 1 close?

pontusjo (approved):

dao/src/main/java/se/fortnox/reactivewizard/db/ReactiveStatementFactory.java
```diff
@@ -53,29 +56,34 @@ private static void closeSilently(Connection connection) {
...
+
+    /**
+     * Create Flux statement
+     *
+     * @param args                the arguments
+     * @param connectionScheduler the scheduler
+     * @param methodReturnType    the return type of the dao method
+     * @return the Flux statement
+     */
+    public <T extends Publisher<?>> Object create(Object[] args, ConnectionScheduler connectionScheduler, Class<T> methodReturnType) {
```
JonasHallFnx, line 84: The methodReturnType is already considered when `resultConverter` is created at startup. Can the new logic be added at startup as well? Java refelection is not very performant so it is nice to avoid reflection in runtime, like below.

dao/src/main/java/se/fortnox/reactivewizard/db/ReactiveStatementFactory.java
```diff
@@ -85,6 +93,19 @@ public Object create(Object[] args, ConnectionScheduler connectionScheduler) {
             });
         }
 
+        // On the possibility that FluxSink signals next and then an
```
JonasHallFnx, line 96: If there was an insert, and the database returned `1` rows affected, but the connection could not be closed after that due to network error or similar, then the new behaviour would be to return an error? Wouldn't that be wrong? The caller would assume that an error means that the insert was not successful?

I may have misunderstood this. Cannot find any new test that asserts there was an error either.

dao/src/main/java/se/fortnox/reactivewizard/db/DbProxy.java
```diff
@@ -123,7 +129,24 @@ public Object invoke(Object proxy, Method method, Object[] args) throws Throwabl
                     converterFromFlux(method.getReturnType()));
             statementFactories.put(method, reactiveStatementFactory);
         }
-        return reactiveStatementFactory.create(args, connectionScheduler);
+
+        if (isFluxOrMono(method.getReturnType())) {
```
JonasHallFnx, line 133: Avoid reflection at runtime, this check can be done inside the block above, which will only happen once per method.

dao/src/main/java/se/fortnox/reactivewizard/db/ReactiveStatementFactory.java
JonasHallFnx: I suggest to never return an error for connection problems that happens after all data has been returned. Connections are pooled, so you could argue that the closing of the connection has nothing to do with the query.

Was there any new scenario that caused you to change the behaviour here?

dao/src/main/java/se/fortnox/reactivewizard/db/DbProxy.java
```diff
@@ -113,19 +119,38 @@ public Object invoke(Object proxy, Method method, Object[] args) throws Throwabl
                 // Need to get the actual interface method in order to get updated annotations
                 method = ReflectionUtil.getRedefinedMethod(method);
             }
+
+            if (!isFluxOrMono(method.getReturnType())) {
+                LOG.error("DAO method {}::{} must return a Flux or Mono. Found {}",
```
noxreview, line 124: I think the exception is enough since that will be logged as error also.

dao/src/main/java/se/fortnox/reactivewizard/db/DbProxy.java
```diff
@@ -113,19 +119,38 @@ public Object invoke(Object proxy, Method method, Object[] args) throws Throwabl
...
-                    converterFromFlux(method.getReturnType()));
+                    converterFromFlux(method.getReturnType()),
+                    method.getReturnType());
             statementFactories.put(method, reactiveStatementFactory);
         }
+
         return reactiveStatementFactory.create(args, connectionScheduler);
     }
 
+    private boolean isFluxOrMono(Class<?> type) {
```
noxreview, line 150: This can be static.

dao/src/main/java/se/fortnox/reactivewizard/db/MonoFluxSink.java
```diff
@@ -0,0 +1,92 @@
...
+import reactor.core.publisher.FluxSink;
+import reactor.util.context.Context;
+import reactor.util.context.ContextView;
+
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.function.LongConsumer;
+
+import static java.util.Objects.nonNull;
+
+public class MonoFluxSink<T> implements FluxSink<T> {
```
flowertwig, line 13: Javadoc

dao/src/main/java/se/fortnox/reactivewizard/db/MonoFluxSink.java
```diff
@@ -0,0 +1,92 @@
...
+        this.isMono = isMono;
+        this.numNextSignals = new AtomicLong(1);
+    }
+
+    @Override
+    public FluxSink<T> next(T t) {
+        if (isMono) {
+            long numSignals = numNextSignals.getAndIncrement();
+            if (numSignals > 1) {
+                actual.error(new RuntimeException("Attempting to signal next multiple times on a Mono"));
```
flowertwig, line 31: Maybe this message is a bit too low-level. We could add a short sentence about the most likely cause.

dao/src/main/java/se/fortnox/reactivewizard/db/ReactiveStatementFactory.java
```diff
@@ -31,20 +34,28 @@ public class ReactiveStatementFactory {
     private final PublisherMetrics publisherMetrics;
 
     private final DatabaseConfig config;
+    private boolean isReturnTypeMono = false;
```
JonasHallFnx, line 37: all final

dao/src/main/java/se/fortnox/reactivewizard/db/DbProxy.java
```diff
@@ -113,19 +117,37 @@ public Object invoke(Object proxy, Method method, Object[] args) throws Throwabl
...
         return reactiveStatementFactory.create(args, connectionScheduler);
     }
 
+    private static Function<Flux, Object> converterFromFlux(Method method) {
+        Class<?> returnType = method.getReturnType();
+
+        if (Flux.class.isAssignableFrom(returnType)) {
+            return flux -> flux;
+        } else if (Mono.class.isAssignableFrom(returnType)) {
+            return Mono::from;
```
JonasHallFnx, line 142: Replace with `Flux::.singleOrEmpty` and drop class `DaoFluxSink`?

dao/src/main/java/se/fortnox/reactivewizard/db/ReactiveStatementFactory.java
JonasHallFnx: These fields can be dropped if DaoFluxSink is dropped

softqwewasd:
- [ ] squash commits if all good

dao/src/main/java/se/fortnox/reactivewizard/db/DbProxy.java
softqwewasd: We lose other important behavior, such as not sending the error if element has been emitted. I've tried to find a one-liner solution but I could not find such solution. Using MonoSink would work but we would be moving code from the custom sink into the MonoSink.

JonasHallFnx (approved):

dao/src/main/java/se/fortnox/reactivewizard/db/ReactiveStatementFactory.java
```diff
@@ -53,29 +67,34 @@ private static void closeSilently(Connection connection) {
         }
     }
 
+    private Flux<Object> getResult(StatementContext statementContext) {
+        return Flux.create(fluxSink -> {
+            var monoFluxSink = new DaoFluxSink<>(fluxSink, isReturnTypeMono, methodName);
```
flowertwig, line 72: variable name should be changed to reflect class name.

flowertwig (approved):

pontusjo (approved):

flowertwig (approved):

mw223bc (approved):

PR #419, opened 2022-10-20
Migrated the test module from RxJava to Reactor

softqwewasd:


softqwewasd:
> I'm thinking that maybe there is not much value in creating reactor versions of ObservableAssert etc. StepVerifier should be sufficient and it also more accurately reflects how Publishers work.

I agree, is there any use of WaitingTestUtil when there is Awaitility? I'm thinking that we can just delete the entire module.

softqwewasd:
> I'm thinking that maybe there is not much value in creating reactor versions of ObservableAssert etc. StepVerifier should be sufficient and it also more accurately reflects how Publishers work.

I agree, do we even need WaitingTestUtils when there is Awaitility?

flowertwig (approved):

PR #425, opened 2022-11-01
Timeout of http client is now configurable

softqwewasd:


client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
```diff
@@ -119,6 +119,7 @@ public HttpClient(HttpClientConfig config,
         collector            = new ByteBufCollector(config.getMaxResponseSize());
         this.preRequestHooks = preRequestHooks;
         this.retryDuration   = Duration.ofMillis(config.getRetryDelayMs());
+        this.timeout         = config.getReadTimeoutMs();
```
flowertwig, line 122: This will be a change of functionality since readTimeoutMs is now used for this as well. We need to check that it doesn't lead to unexpected behaviour for existing configurations.

client/src/test/java/se/fortnox/reactivewizard/client/HttpClientTest.java
```diff
@@ -1442,7 +1442,7 @@ public void shouldNotRetryFailedPostCallsWithStrangeExceptions() {
         DisposableServer server    = startServer(INTERNAL_SERVER_ERROR, "\"NOT OK\"", r -> callCount.incrementAndGet());
         try {
             HttpClientConfig config = new HttpClientConfig("localhost:" + server.port());
-            config.setReadTimeoutMs(100);
+            config.setReadTimeoutMs(500);
```
flowertwig, line 1445: Why change this? There are no changes to assertions or similar.

flowertwig (changes requested):

client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
softqwewasd: I thought that the `readTimeoutMs` is the same as `HttpClient::timeout` since we are checking for a `ReadTimeoutException` in [HttpClient::convertError](https://github.com/FortnoxAB/reactive-wizard/blob/c6a2b1026d833b322f237e4ba6fd4f2a21300b5c/client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java#L287) and sending an error with the HttpClient::timeout when it should be `readTimeoutMs`. In other words, the message in the `WebException` when a `ReadTimeoutException` is thrown would not be correct when `readTimeoutMs` is changed to some other value than 10k ms. Either `readTimeoutMs` is the same as `HttpClient::timeout` or the handling of `ReadTimeoutException` and `TimeoutException` is wrong.

PR #440, opened 2022-12-11
Make it easier to use custom classes as request parameters

TommyPersson:
Follows the JAX-RS spec by allowing classes with:
* A static `valueOf(String)`-method
* A static `fromString(String)`-method
* A `String`-constructor

### DIFF
--- a/jaxrs/src/main/java/se/fortnox/reactivewizard/jaxrs/params/deserializing/CustomClassDeserializerFactory.java
+++ b/jaxrs/src/main/java/se/fortnox/reactivewizard/jaxrs/params/deserializing/CustomClassDeserializerFactory.java
@@ -6,6 +6,10 @@ import java.lang.reflect.Modifier;
 import java.util.Arrays;
 
 class CustomClassDeserializerFactory {
+
+    private CustomClassDeserializerFactory() {
+    }
+
     @Nullable
     public static <T> Deserializer<T> createOrNull(Class<T> cls) {
         var methods = cls.getMethods();
--- a/jaxrs/src/main/java/se/fortnox/reactivewizard/jaxrs/params/deserializing/CustomClassDeserializerFactory.java
+++ b/jaxrs/src/main/java/se/fortnox/reactivewizard/jaxrs/params/deserializing/CustomClassDeserializerFactory.java
@@ -1,30 +1,37 @@
 package se.fortnox.reactivewizard.jaxrs.params.deserializing;
 
+import se.fortnox.reactivewizard.util.LambdaCompiler;
+
 import javax.annotation.Nullable;
+import java.lang.invoke.MethodHandles;
+import java.lang.reflect.Constructor;
 import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
 import java.util.Arrays;
+import java.util.function.Function;
 
 class CustomClassDeserializerFactory {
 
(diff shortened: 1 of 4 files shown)


jaxrs/src/main/java/se/fortnox/reactivewizard/jaxrs/params/deserializing/DeserializerFactory.java
```diff
@@ -18,87 +19,109 @@
...
+        var paramCls = ReflectionUtil.getRawType(paramType.getType());
 
         if (List.class.isAssignableFrom(paramCls)) {
-            Class<?>     elementType = ReflectionUtil.getGenericParameter(paramType.getType());
+            var          elementCls = ReflectionUtil.getGenericParameter(paramType.getType());
             Deserializer elementDeserializer;
 
-            if (elementType.isEnum()) {
-                elementDeserializer = new EnumDeserializer(elementType);
+            if (elementCls.isEnum()) {
```
JonasHallFnx, line 72: Can this be skipped now since you will do this at line 102 now?

jaxrs/src/main/java/se/fortnox/reactivewizard/jaxrs/params/deserializing/CustomClassDeserializerFactory.java
```diff
@@ -0,0 +1,92 @@
...
+            .orElse(null);
+
+        if (method != null) {
+            return value -> {
+                try {
+                    if (value == null) {
+                        return null;
+                    }
+
+                    return (T)method.invoke(null, value);
```
JonasHallFnx, line 54: Perhaps you can use `LambdaCompiler.compileLambdaFunction` to avoid reflection at runtime? Applies below as well.

TommyPersson: I'll have a look.

jaxrs/src/main/java/se/fortnox/reactivewizard/jaxrs/params/deserializing/CustomClassDeserializerFactory.java
```diff
@@ -0,0 +1,92 @@
...
+class CustomClassDeserializerFactory {
+
+    private CustomClassDeserializerFactory() {
+    }
+
+    @Nullable
+    public static <T> Deserializer<T> createOrNull(Class<T> cls) {
+        var methods = cls.getMethods();
+
+        var valueOfDeserializer = findStaticMethodDeserializer(methods, "valueOf", cls);
```
flowertwig: Optional can be used to skip the nullchecks here if you wish, but it's not necessary.

```
return ofNullable(findStaticMethodDeserializer(methods, "valueOf", cls))
			.or(() -> ofNullable(findStaticMethodDeserializer(methods, "fromString", cls)))
			.or(() -> ofNullable(findConstructorDeserializer(cls)))
			.orElse(null);
```

TommyPersson: I'll let you decide. I personally slightly prefer the null-checks in this scenario.

JonasHallFnx (approved):

softqwewasd (approved):

PR #445, opened 2022-12-13
Added MockitoExtension dependency

flowertwig:
Added MockitoExtension dependency in test module to be able to use `@Mock` annotation in JUnit5 tests.
Added example of usage in FluxClientTest.
Removed superfluous dependency management for reactive-streams.
Updated description in root pom.

### DIFF
--- a/test/pom.xml
+++ b/test/pom.xml
@@ -63,8 +63,8 @@
         </dependency>
 
         <dependency>
-            <groupId>org.slf4j</groupId>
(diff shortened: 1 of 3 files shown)


test/pom.xml
```diff
@@ -63,8 +63,8 @@
         </dependency>
 
         <dependency>
-            <groupId>org.slf4j</groupId>
```
flowertwig, line 66: This dependency was a duplicate.

PR #456, opened 2023-01-03
Migrated RetryWithDelay from RxJava to Projector

softqwewasd:


utils/src/main/java/se/fortnox/reactivewizard/util/rx/RetryWithDelay.java
```diff
@@ -1,48 +1,44 @@
 package se.fortnox.reactivewizard.util.rx;
```
flowertwig, line 1: I like Project Reactor -> Projector :D 
These changes will make it unusable for rxjava and for projector it seems pretty unnecessary. I suggest we either remove it or keep it until rxjava support is gone.

flowertwig (changes requested):

utils/src/main/java/se/fortnox/reactivewizard/util/rx/RetryWithDelay.java
softqwewasd: Agreed! Closing PR.

PR #457, opened 2023-01-09
Updated slf4j artifact

flowertwig:
Changed to more recent slf4j implementation artifact and bumped version.
Updated some old versions and artifacts in the readme.


softqwewasd (approved):
nice!

PR #465, opened 2023-01-19
Implemented SQL prepared statements logging

softqwewasd:


dao/src/main/java/se/fortnox/reactivewizard/db/statement/StatementDebug.java
```diff
@@ -0,0 +1,27 @@
+package se.fortnox.reactivewizard.db.statement;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.sql.PreparedStatement;
+
+public class StatementDebug {
+    private static final Logger LOG = LoggerFactory.getLogger(StatementDebug.class);
+    static boolean ENABLE_STATEMENT_DEBUG = Boolean.parseBoolean(System.getenv("dao.debug"));
```
JonasHallFnx, line 10: Can this be final so that the if-statement below cannot change at runtime and the JVM can "remove" the code?

JonasHallFnx (approved):

dao/src/main/java/se/fortnox/reactivewizard/db/statement/StatementDebug.java
```diff
@@ -0,0 +1,27 @@
...
+    }
+
+    /**
+     * Logging statement if dao.debug is set to true.
+     * This only works if an actual postgres database
+     * is running.
+     * @param statement a sql PreparedStatement
+     */
+    public static void log(PreparedStatement statement) {
+        if (ENABLE_STATEMENT_DEBUG) {
```
jepp3, line 22: Do you know if there is a reason for using an environment variable here? 

Otherwise, I think it would be easier to go with the standard way by configuring the logging through the admin port on demand or in the log4j.xml file.

Also Log4j2 support lazy logging with lambdas like this:

`LOG.debug("Executed statement: {}", () -> statement);`

If you do not want to create the string "Executed statement: {}" every time due to optimization you can put it in a final static field in the class as well.

jepp3 (approved):

dao/src/main/java/se/fortnox/reactivewizard/db/statement/StatementDebug.java
softqwewasd: We have to be able to change the flag in the tests. Unsure if Mockito has a nice api for that. Perhaps the value of the field can be changed with reflection but I don't think that is an OK solution to this problem.  We could create a static method, getEnableStatementDebug, that returns the ENABLE_STATEMENT_DEBUG, which can then be mocked by enabling Mockito's inline mock maker. That would solve one part of the problem, not sure though if JVM will inline the log method calls when you call getEnableStatementDebug instead of just checking the ENABLE_STATEMENT_DEBUG field.

dao/src/main/java/se/fortnox/reactivewizard/db/statement/StatementDebug.java
JonasHallFnx: Making it dynamic means you would get a penalty on each call because the flag needs to be checked. If it can be set using static+final then the JVM can completely remove this code when it is compiled.

softqwewasd: It is not meant to be an environment variable, it should be a java property, my bad, will fix.

dao/src/main/java/se/fortnox/reactivewizard/db/statement/StatementDebug.java
```diff
@@ -0,0 +1,27 @@
...
+            LOG.info("Executed statement: {}", statement);
```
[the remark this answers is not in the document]
softqwewasd: Yes. it seems like postgres's jdbc driver is one of the few that will actually print the prepared statement with the actual values.

dao/src/main/java/se/fortnox/reactivewizard/db/statement/StatementDebug.java
JonasHallFnx: Why do you need to change the flag in the tests? Can't you just decide that all tests are run with the flag on or off?

softqwewasd: I was thinking specifically of [these tests](https://github.com/FortnoxAB/reactive-wizard/pull/465/commits/28ec8be3a02506ac66dc3655b012fc049f98071a#diff-81166375d5db84e39cfb0e42b65f7c220291d8d7dab049420b4845c768cf10cb). I have manually tested these cases instead.

JonasHallFnx: Personally, I would rather have the statement final than having it possible to automaticall test the behaviour when it is off.

flowertwig (approved):

PR #471, opened 2023-01-31
Feature/stream format

softqwewasd:
**Summary of this PR**

1. `Flux<T>` resource methods no longer streaming by default
2. `Stream` annotation is now dependent on a parameter that determines the response body format of the stream.
    - `Stream.Type.CONCATENATED_JSON_OBJECTS` is equivalent to the format used with `@Stream Observable<T>` resources
    - `Stream.Type.JSON_ARRAY` is the same as the current `Flux<T>` stream. This is the default stream format.
3. `HttpClient` can now deserialize an `Observable<T>` stream.
4. The `@Stream` annotation should now be annotated on the interface level.

**(1) `Flux<T>` resource method no longer streaming by default**
A chunked HTTP response is not the same as a non-chunked response. When the response is chunked the status code is sent on the first element. The status code cannot be changed afterwards. This has lead to people migrating their non-streaming `Observable` resources to streaming Flux resources which has caused issues to its dependents. Here is a common scenario:

```
@Path("/api/serviceA")
@GET
Flux<Entity> getMyEntities()
```

```
@Path("api/serviceB")
@GET
public Flux<Message> getMessages() {
     messageDao.getMessages().concatMap(message -> {
        return serviceA.getMyEntities().concatMap(entity -> {
            if (someLogic(entity, message)) {
                return doSomethingAndReturnMessage(entity, message);
            }
            return Flux.error(new RuntimeException("Something went wrong!")) 
        })
    })
}
```
In this scenario, the data sent from _service A_ may be fine, but there are some business logic in _service B_ that may end up error signaling in the middle of the stream. If the first signal is not an error signal in `getMessages` then the error is rendered useless since the client will never receive it. 

This change is backwards compatible. There is no way to resolve the issue in the code by leaving `Flux<T>` as streaming by default and also keep backwards compatibility.

**(2) `Stream` annotation is now dependent on a parameter that determines the response body format of the stream.**
A streaming `Observable<T>` resource method is not the same as a streaming `Flux<T>` resource method. The difference between the two is their serialization. Each item from an `Observable<T>` is serialized and sent to the client. The client will essentially have a string of concatenated JSON objects: `{"message": "Hello"}{"message": "World"}`. The serialization in a `Flux<T>` stream is a JSON array: `[{"message": "Hello"}, {"message": "World"}]`.

If a service has a streaming `Observable` resource method, the migration to a streaming `Flux` will cause sudden deserialization errors to the HTTP clients.

The solution here is to support both serializations.

**(3) `HttpClient` can now deserialize an `Observable<T>` stream.**
Current RW clients cannot deserialize a stream of concatenated json objects. Since we are going to continue to support the serialization, we should support the deserialization as well. Furthermore, if there are people that have written their own HTTP clients to support the deserialization, then those clients can now be replaced with the RW client.

**(4) The `@Stream` annotation should now be annotated on the interface level.**
While we can introduce a custom content-type(ex. application/cojson) to separate the concatenated json object format and the json array format, we should not change that now since it may cause issues to clients that check the content-type against application/json(ex. in the UI:s). In order for the RW client to separate the two stream formats, the annotation must be on the resource api and not the implementation.


client/src/main/java/se/fortnox/reactivewizard/client/ConcatenatedJsonObjectsDeserializer.java
```diff
@@ -0,0 +1,123 @@
...
+import java.util.ArrayDeque;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Deque;
+import java.util.LinkedList;
+import java.util.List;
+
+import static java.lang.String.format;
+
+public class ConcatenatedJsonObjectsDeserializer {
```
magnusp, line 27: Have we looked at using https://www.javadoc.io/static/com.fasterxml.jackson.core/jackson-core/2.14.2/com/fasterxml/jackson/core/JsonParser.html?

client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java
```diff
@@ -317,10 +326,18 @@ private String resolveContentType(Method method, RwHttpClientResponse response)
         return contentType;
     }
 
-    protected Flux<Object> parseResponseStream(Method method, RwHttpClientResponse response) {
+    protected Flux<Object> parseResponseStream(Method method, RwHttpClientResponse response, JaxRsMeta jaxRsMeta) {
         String contentType = resolveContentType(method, response);
+        Optional<Stream> streamAnnotation = jaxRsMeta.getStream();
+        if (streamAnnotation.isPresent() && CONCATENATED_JSON_OBJECTS.equals(streamAnnotation.get().value())
+            || "application/cojson".contentEquals(contentType)) {
```
magnusp, line 333: cojson -> x-cojson

https://www.w3.org/Protocols/rfc1341/4_Content-Type.html
> Private values (starting with "X-") may be defined bilaterally between two cooperating agents without outside registration or standardization.

magnusp:
Have we done any browser based manual testing in regard to the new content-type? How does curl behave?

client/src/main/java/se/fortnox/reactivewizard/client/ConcatenatedJsonObjectsDeserializer.java
```diff
@@ -0,0 +1,123 @@
...
+                tokenStack.pop();
+            } else if (isCitation(currentByte) && isCitation(tokenStack.peek()) && !isEscaped()) {
+                tokenStack.pop();
+            } else if (isCitation(currentByte) && !isCitation(tokenStack.peek()) && !isEscaped()) {
+                tokenStack.push(currentByte);
+            }
+
+            if (tokenStack.isEmpty()) {
+                contentToAppend = Arrays.copyOfRange(responseBodyChunk, startParsingFrom, i + 1);
+                currentJsonObject = currentJsonObject.concat(new String(contentToAppend, StandardCharsets.UTF_8)).trim();
```
magnusp, line 73: We could just push the bytes onto a buffer with a type that fits readValue - that would avoid allocating strings.

client/src/main/java/se/fortnox/reactivewizard/client/ConcatenatedJsonObjectsDeserializer.java
jepp3, line 27: I dont see any reason to introduce support for the client to handle the old streaming {obj},{obj} format. I think we should avoid to add logic to the framework that we actually do not need. If we would add a serialization support in the rw client I think it would be for a format that we do not think is a legacy format.

jaxrs-api/src/main/java/se/fortnox/reactivewizard/jaxrs/Stream.java
```diff
@@ -11,5 +11,18 @@
...
+     * then the format will be just like an Observable stream
+     * resource method. If it is set to JSON_ARRAY then it will
+     * wrap the stream in a JSON array. This option applies only
+     * for the resource methods returning Flux.
+     * @return a stream type
+     */
+    Type value() default Type.JSON_ARRAY;
 
+    enum Type {
+        CONCATENATED_JSON_OBJECTS,
```
jepp3, line 25: Can you somehow mark concatenated_json_objects as deprected?

jaxrs-api/src/main/java/se/fortnox/reactivewizard/jaxrs/Stream.java
```diff
@@ -11,5 +11,18 @@
 @Target({ElementType.METHOD})
 @Retention(RetentionPolicy.RUNTIME)
 public @interface Stream {
+    /**
+     * Stream format. If it set to CONCATENATED_JSON_OBJECTS,
```
jepp3, line 15: The reader may not know what format you are referring to when you write _"if it set to CONCATENATED_JSON_OBJECTS,  then the format will be just like an Observable stream"_

maybe more valuable if you state how the format will be.

jaxrs/src/main/java/se/fortnox/reactivewizard/jaxrs/response/JaxRsResultFactory.java
```diff
@@ -80,10 +101,10 @@ public JaxRsResult<T> create(Flux<T> output, Object[] args) {
         return result;
     }
 
-    private Func1<Flux<T>, Flux<byte[]>> createSerializer(String type, Class<T> dataCls, boolean isSingleType, JsonSerializerFactory jsonSerializerFactory) {
+    private Func1<Flux<T>, Flux<byte[]>> createStreamSerializer(String type, Class<T> dataCls, Type streamType, JsonSerializerFactory jsonSerializerFactory) {
```
jepp3, line 104: We have two JaxRsResultFactories. The one we have here and also the one that is named [JaxRsStreamingResultFactory](https://github.com/FortnoxAB/reactive-wizard/blob/master/jaxrs/src/main/java/se/fortnox/reactivewizard/jaxrs/response/JaxRsStreamingResultFactory.java). Why do we have the code that manage the streaming in the parent class rather in the child class with the naming "streaming"?

client/src/main/java/se/fortnox/reactivewizard/client/ConcatenatedJsonObjectsDeserializer.java
softqwewasd: I did. Looking at [JsonArrayDeserializer](https://github.com/FortnoxAB/reactive-wizard/blob/master/client/src/main/java/se/fortnox/reactivewizard/client/JsonArrayDeserializer.java), I thought it may be easier to just write the deserializer without using JsonParser :)

client/src/main/java/se/fortnox/reactivewizard/client/ConcatenatedJsonObjectsDeserializer.java
softqwewasd: Good point. Perhaps we could have gotten a better solution here if JsonParser was used. That said, I deleted ConcatenatedJsonObjectDeserializer alltogether.

client/src/main/java/se/fortnox/reactivewizard/client/ConcatenatedJsonObjectsDeserializer.java
softqwewasd: Agreed!

jaxrs/src/main/java/se/fortnox/reactivewizard/jaxrs/JaxRsResultSerializerFactory.java
```diff
@@ -0,0 +1,92 @@
...
+            }
+        }
+        if (dataCls.equals(BYTEARRAY_TYPE)) {
+            return flux -> flux.cast(byte[].class);
+        }
+
+        if (isFlux) {
+            return flux -> flux.map(Object::toString).reduce(String::concat).map(String::getBytes).flux();
+        }
+        Mono.empty().switchIfEmpty(Mono.error(new RuntimeException("")));
```
jepp3, line 52: I guess this one is just fortgotten here and should be removed?

magnusp (approved):

jaxrs/src/main/java/se/fortnox/reactivewizard/jaxrs/JaxRsResultSerializerFactory.java
```diff
@@ -0,0 +1,92 @@
...
+
+    /**
+     * Creates a non streaming result serializer
+     * @param type Content-Type
+     * @param dataCls Return type of the resource method
+     * @param isFlux True if it is a flux, false otherwise
+     * @param <T> Type of elements emitted by the publisher
+     * @return a function that serializes the elements emitted by a publisher to a Flux of byte arrays.
+     */
+    public <T> Func1<Flux<T>, Flux<byte[]>> createNonStreamingSerializer(String type, Class<T> dataCls, boolean isFlux) {
```
flowertwig, line 37: Can't we use an ordinary java util functions mostly everywhere?

softqwewasd:
- [ ] squash

jaxrs/src/main/java/se/fortnox/reactivewizard/jaxrs/response/JaxRsResultFactory.java
```diff
@@ -116,7 +83,7 @@ private Class<T> getRawReturnType(Method method) {
         return (Class<T>)method.getReturnType();
     }
 
-    private HttpResponseStatus getSuccessStatus(JaxRsResource resource) {
+    protected HttpResponseStatus getSuccessStatus(JaxRsResource resource) {
```
flowertwig, line 86: Not needed?

jaxrs/src/main/java/se/fortnox/reactivewizard/jaxrs/response/JaxRsResultFactoryFactory.java
```diff
@@ -29,12 +31,35 @@ public JaxRsResultFactoryFactory(ResultTransformerFactories resultTransformerFac
...
-        boolean isSingleType = FluxRxConverter.isSingleType(resource.getInstanceMethod().getReturnType());
-        if (resource.getInstanceMethod().isAnnotationPresent(Stream.class) || !isSingleType) {
-            return new JaxRsStreamingResultFactory<>(resource, resultTransformerFactories, jsonSerializerFactory);
+    private <T> boolean isStream(JaxRsResource<T> resource) {
+        return resource.getInstanceMethod().isAnnotationPresent(Stream.class);
+    }
+
+    private <T> boolean isFluxByteArray(JaxRsResource<T> resource) {
+        return resource.getInstanceMethod().getReturnType().isAssignableFrom(Flux.class)
+            && resource.getProduces().equals("application/octet-stream");
```
flowertwig, line 54: There is a MediaType with this as a constant if we'd rather use that.

flowertwig (changes requested):

jaxrs/src/main/java/se/fortnox/reactivewizard/jaxrs/JaxRsResultSerializerFactory.java
softqwewasd: Since it was an Func1 before my changes I thought I'd leave it as is. That said, I do agree that it should be a java util function rather than an rx function.

flowertwig (approved):

jaxrs/src/main/java/se/fortnox/reactivewizard/jaxrs/JaxRsResultSerializerFactory.java
```diff
@@ -0,0 +1,90 @@
...
+     * Creates a non streaming result serializer
+     * @param type Content-Type
+     * @param dataCls Return type of the resource method
+     * @param isFlux True if it is a flux, false otherwise
+     * @param <T> Type of elements emitted by the publisher
+     * @return a function that serializes the elements emitted by a publisher to a Flux of byte arrays.
+     */
+    public <T> Function<Flux<T>, Flux<byte[]>> createNonStreamingSerializer(String type, Class<T> dataCls, boolean isFlux) {
+        if (type.equals(MediaType.APPLICATION_JSON)) {
+            if (!isFlux) {
```
jepp3, line 38: what do you think about renaming isFlux to returnTypeIsFlux to make it more clear? IsFlux can mean a lot if you do not have the context.

jaxrs/src/main/java/se/fortnox/reactivewizard/jaxrs/JaxRsResultSerializerFactory.java
```diff
@@ -0,0 +1,90 @@
...
+
+    /**
+     * Creates a non streaming result serializer
+     * @param type Content-Type
+     * @param dataCls Return type of the resource method
+     * @param isFlux True if it is a flux, false otherwise
+     * @param <T> Type of elements emitted by the publisher
+     * @return a function that serializes the elements emitted by a publisher to a Flux of byte arrays.
+     */
+    public <T> Function<Flux<T>, Flux<byte[]>> createNonStreamingSerializer(String type, Class<T> dataCls, boolean isFlux) {
```
jepp3, line 36: I dont think you need to write "non" in the name. Maybe just call it createSerializer

jaxrs/src/main/java/se/fortnox/reactivewizard/jaxrs/JaxRsResultSerializerFactory.java
softqwewasd: Yes, good point. That is more clear.

jaxrs/src/main/java/se/fortnox/reactivewizard/jaxrs/response/JaxRsResultFactoryFactory.java
```diff
@@ -29,12 +33,35 @@ public JaxRsResultFactoryFactory(ResultTransformerFactories resultTransformerFac
...
-            return new JaxRsStreamingResultFactory<>(resource, resultTransformerFactories, jsonSerializerFactory);
+        if (isStream(resource)) {
+            return new JaxRsStreamingResultFactory<>(resource, resultTransformerFactories, jaxRsResultSerializerFactory);
         }
 
-        return new JaxRsResultFactory<>(resource, resultTransformerFactories, jsonSerializerFactory);
+        return new JaxRsResultFactory<>(resource, resultTransformerFactories, jaxRsResultSerializerFactory);
+    }
+
+    private <T> boolean isStream(JaxRsResource<T> resource) {
```
jepp3, line 50: isStreamAnnotationPresent?

jaxrs/src/main/java/se/fortnox/reactivewizard/jaxrs/JaxRsResultSerializerFactory.java
```diff
@@ -0,0 +1,90 @@
...
+     * @param streamType Stream type
+     * @param isFlux True if it is a flux, false otherwise
+     * @param <T> Type of elements emitted by the publisher
+     * @return a function that serializes the elements emitted by a publisher to a Flux of byte arrays.
+     */
+    public <T> Function<Flux<T>, Flux<byte[]>> createStreamingSerializer(String type, Class<T> dataCls, Stream.Type streamType, boolean isFlux) {
+        if (type.equals(MediaType.APPLICATION_JSON)) {
+            var byteSerializer = jsonSerializerFactory.createByteSerializer(dataCls);
+            if (streamType == Stream.Type.CONCATENATED_JSON_OBJECTS || !isFlux) {
+                return flux -> flux.map(byteSerializer);
```
jepp3, line 67: It is a flux of course, but what does it contain?  Maybe its better to call it items? 

items -> items.map(byteSerilizer)

jaxrs/src/main/java/se/fortnox/reactivewizard/jaxrs/JaxRsResultSerializerFactory.java
```diff
@@ -0,0 +1,90 @@
...
+            } else {
+                return flux -> {
+                    AtomicBoolean first = new AtomicBoolean(true);
+                    Flux<byte[]> items = flux.concatMap(item -> {
```
jepp3, line 71: Same here, but maybe it should be called serializedItems

jaxrs-api/src/main/java/se/fortnox/reactivewizard/jaxrs/Stream.java
JonasHallFnx: Why are we _adding_ an option that we mark as deprecated...?

jaxrs/src/main/java/se/fortnox/reactivewizard/jaxrs/JaxRsResultSerializerFactory.java
```diff
@@ -0,0 +1,90 @@
...
+     * @param type Content-Type
+     * @param dataCls Return type of the resource method
+     * @param isFlux True if it is a flux, false otherwise
+     * @param <T> Type of elements emitted by the publisher
+     * @return a function that serializes the elements emitted by a publisher to a Flux of byte arrays.
+     */
+    public <T> Function<Flux<T>, Flux<byte[]>> createNonStreamingSerializer(String type, Class<T> dataCls, boolean isFlux) {
+        if (type.equals(MediaType.APPLICATION_JSON)) {
+            if (!isFlux) {
+                return flux -> flux.map(jsonSerializerFactory.createByteSerializer(dataCls));
```
JonasHallFnx, line 39: move creation of serializer outside of lambda to avoid that a new serializer is created for each request?

jaxrs/src/main/java/se/fortnox/reactivewizard/jaxrs/JaxRsResultSerializerFactory.java
```diff
@@ -0,0 +1,90 @@
...
+            } else {
+                return flux -> flux.buffer().defaultIfEmpty(Collections.emptyList()).map(jsonSerializerFactory.createByteSerializer(new TypeReference<>() { }));
```
JonasHallFnx, line 41: Can you instead create a type that is `List<dataCls>`?

jaxrs/src/main/java/se/fortnox/reactivewizard/jaxrs/JaxRsResultSerializerFactory.java
```diff
@@ -0,0 +1,90 @@
...
+                    AtomicBoolean first = new AtomicBoolean(true);
+                    Flux<byte[]> items = flux.concatMap(item -> {
+                        if (first.getAndSet(false)) {
+                            return just(byteSerializer.apply(item));
+                        } else {
+                            return just(",".getBytes(), byteSerializer.apply(item));
+                        }
+                    });
+                    return Flux.concat(
+                        just("[".getBytes()),
```
JonasHallFnx, line 79: put in static final fields

jaxrs-api/src/main/java/se/fortnox/reactivewizard/jaxrs/Stream.java
jepp3: For allowing backward compatibility. Some services are currently producing the "concatenated JSON format" and converting the endpoint to reactor will change the format, resulting in breaking changes. adding concatenated_json_object to the endpoint stream annotation will allow clients to consume the data as usual. Is just a way to mark the old way of streaming data.

If the concatenated json shall be deprecated may of course be discussed I think. 

Would you rather see it as a nondeprecated format?

jaxrs/src/main/java/se/fortnox/reactivewizard/jaxrs/JaxRsResultSerializerFactory.java
```diff
@@ -0,0 +1,90 @@
+package se.fortnox.reactivewizard.jaxrs;
```
JonasHallFnx, line 1: Perhaps put this class in the `response` package with the other associated classes?

jaxrs-api/src/main/java/se/fortnox/reactivewizard/jaxrs/Stream.java
JonasHallFnx: Ok, thanks for explaining :-)

jaxrs/src/main/java/se/fortnox/reactivewizard/jaxrs/JaxRsResultSerializerFactory.java
jepp3: Good catch @JonasHallFnx :1st_place_medal: . You may have saved us from unwanted stress-related baldness.

jaxrs/src/main/java/se/fortnox/reactivewizard/jaxrs/JaxRsResultSerializerFactory.java
softqwewasd: I could not get a List<dataCls>, not sure if that is possible :thinking:

jepp3 (approved):

jaxrs/src/main/java/se/fortnox/reactivewizard/jaxrs/JaxRsResultSerializerFactory.java
JonasHallFnx: `objectMapper.getTypeFactory().constructCollectionType(List.class, dataCls)`

jaxrs/src/main/java/se/fortnox/reactivewizard/jaxrs/response/JaxRsResultSerializerFactory.java
```diff
@@ -0,0 +1,100 @@
...
+     * Creates a non streaming result serializer
+     * @param type Content-Type
+     * @param dataCls Return type of the resource method
+     * @param returnTypeIsFlux True if it is a flux, false otherwise
+     * @param <T> Type of elements emitted by the publisher
+     * @return a function that serializes the elements emitted by a publisher to a Flux of byte arrays.
+     */
+    public <T> Function<Flux<T>, Flux<byte[]>> createSerializer(String type, Class<T> dataCls, boolean returnTypeIsFlux) {
+        var byteSerializer = jsonSerializerFactory.createByteSerializer(dataCls);
+        var jsonSerializer = jsonSerializerFactory.createByteSerializer(new TypeReference<>() { });
```
JonasHallFnx, line 44: `objectMapper.getTypeFactory().constructCollectionType(List.class, dataCls)`

JonasHallFnx: Perhaps move these into the correct if-block below to avoid creating serializers that are not needed?

JonasHallFnx (approved):

PR #515, opened 2023-03-29
Metrics migrated from RxJava to Reactor

softqwewasd:


metrics/src/main/java/se/fortnox/reactivewizard/metrics/Metrics.java
flowertwig, line 1: Should we add methods for Mono as well or try to do something more generic with Publisher?

flowertwig: I missed PublisherMetrics. :P Do we need this then or should we rename PublisherMetrics to Metrics and replace it?

softqwewasd: I think we should have support for Mono and Flux. Otherwise, developers will likely have to convert the Publisher into a Mono or Flux regardless. I think that we can remove the PublisherMetrics.

flowertwig (approved):

PR #525, opened 2023-04-11
Objects contained in iterables are now validated

HarpoMafacka:
Our team noticed an inconsistency between validating single objects and iterables passed as parameters in the API. Consider the following:

_// someObject is validated by default if it has validation annotations_
someResource.create(**SomeObject** object);

_// The same someObject contained in a list is NOT validated by default_
someResource.create(List<**SomeObject**> objects);

_// For iterables, you have to explicitly ask for validation of the list using the @Valid annotation, which will cascade to the objects contained in the iterable_
someResource.create(@Valid List<**SomeObject**> objects);

The risk with this inconsistency is that developers assumes that iterables are validated the same as single objects. This PR contains a suggestion for an implementation that introduces validation of objects contained in iterables as a default behavior.

### DIFF
--- a/validation/src/test/java/se/fortnox/reactivewizard/validation/InputValidationTest.java
+++ b/validation/src/test/java/se/fortnox/reactivewizard/validation/InputValidationTest.java
@@ -411,6 +411,13 @@ public class InputValidationTest {
      * objects) without the need to explicitly add the @Valid annotation
      * to the parameter.
      */
+
+    ValidAcceptingService<Iterable<InputClass>> annotatedService = mock(ValidAcceptingService.class);
+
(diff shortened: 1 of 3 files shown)


validation/src/test/java/se/fortnox/reactivewizard/validation/InputValidationTest.java
flowertwig, line 1: We should probably add a test for the "old" way as well, where the annotation is on the iterable.

HarpoMafacka: I've added two tests to verify that the default validation behavior also applies to input parameters of type Iterable that are annotated with the @Valid annotation.

softqwewasd (approved):

erljef (approved):

PR #534, opened 2023-04-26
Migrated client, server, validation and utils to reactor

softqwewasd:
Observables are still used in the client module and utils module. This is for compatibility reasons and will eventually be changed completely.


utils/src/main/java/se/fortnox/reactivewizard/util/FluxRxConverter.java
```diff
@@ -126,6 +126,9 @@ public static Function<Flux, Object> converterFromFlux(Class<?> returnType) {
      * @param returnType the type to check
      * @return whether the type is reactive
      */
+    // NOTE-TO-SELF: Must still check against Observable and Single since it is used in
```
flowertwig, line 129: Remove when done. :)

tsfcode (approved):

PR #542, opened 2023-05-11
Migrated jaxrs module from RxJava to Reactor.

softqwewasd:


jaxrs/src/main/java/se/fortnox/reactivewizard/jaxrs/response/JaxRsResultFactoryFactory.java
```diff
@@ -32,13 +29,6 @@ public JaxRsResultFactoryFactory(ResultTransformerFactories resultTransformerFac
      * @return the result factory
      */
     public <T> JaxRsResultFactory<T> createResultFactory(JaxRsResource<T> resource) {
-        if (!isStreamAnnotationPresent(resource) && isFluxByteArray(resource)) {
```
softqwewasd, line 35: I believe this was added when we were resolving the compatibility issues and confusion around streaming and non-streaming Flux resources. The removal of this check should not cause any other compatibility issues.

flowertwig (approved):

PR #545, opened 2023-05-16
DAO methods returning Mono<Void> will now complete successfully without emitting a value.

softqwewasd:


tsfcode (approved):

flowertwig (changes requested):
We should have tests for this.

flowertwig (approved):

PR #555, opened 2023-05-31
Changed DAO behavior when MonoSink is involved.

softqwewasd:
Changed DAO behavior when MonoSink is involved such that a call is made to either MonoSink::success(T) or MonoSink::success() but never both.

Despite the fact that a MonoSink::success() should be able to be called multiple times, this resolves an ambiguous backpressure behavior when a DAO method is returned in a concatMap.

### DIFF
--- a/dao/pom.xml
+++ b/dao/pom.xml
@@ -69,6 +69,12 @@
             <classifier>tests</classifier>
         </dependency>
 
+        <dependency>
+            <groupId>org.awaitility</groupId>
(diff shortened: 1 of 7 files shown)


flowertwig (commented):
Some tests would be nice to avoid future regression and clarify the purpose of the changes.

flowertwig (approved):

TommyPersson (approved):

PR #574, opened 2023-07-05
Fixed issue where console log was sometimes disabled in tests

flowertwig:
LoggingMockUtil no longer removes existing appenders when adding a mocked appender. This could sometimes result in console log being lost in tests.

### DIFF
--- a/test/src/main/java/se/fortnox/reactivewizard/test/LoggingMockUtil.java
+++ b/test/src/main/java/se/fortnox/reactivewizard/test/LoggingMockUtil.java
@@ -13,6 +13,7 @@ import org.slf4j.LoggerFactory;
 
 import java.io.Serializable;
 import java.lang.reflect.Field;
+import java.util.List;
 
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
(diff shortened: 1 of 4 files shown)


test/src/main/java/se/fortnox/reactivewizard/test/LoggingMockUtil.java
```diff
@@ -36,35 +37,38 @@ private LoggingMockUtil() {
...
         if (mockAppender != null) {
-            return ((AppenderPreservingEvents)mockAppender).getInner();
+            return ((AppenderPreservingEvents) mockAppender).getInner();
         }
 
         mockAppender = mock(Appender.class);
         when(mockAppender.getName()).thenReturn(MOCK_APPENDER);
+        List<Appender> existingAppenders = List.copyOf(logger.getAppenders().values());
         logger.addAppender(new AppenderPreservingEvents(mockAppender));
+        existingAppenders.forEach(logger::addAppender);
```
softqwewasd, line 53: The console appender is removed when adding the mock appender? :thinking:

tsfcode (approved):

PR #598, opened 2023-09-07
Use ConcurrentHashMap in HttpClientProvider

owtroid:
When creating new httpclients in parallell, sometimes a ConcurrentModificationException would be thrown.


noxreview (approved):

client/src/main/java/se/fortnox/reactivewizard/client/HttpClientProvider.java
```diff
@@ -9,6 +9,7 @@
 import java.util.HashMap;
```
flowertwig, line 9: This is now unused?

owtroid: Yes, removed!

tsfcode (approved):

flowertwig (approved):

PR #631, opened 2023-10-25
Disregard charset in Content-Type when parsing response

flowertwig:
Disregard charset in Content-Type when parsing response, to avoid unnecessary logging if it differs from the default application/json or the Produces annotation on the resource method.


softqwewasd (approved):

jepp3 (approved):

jepp3:
I guess we do not specify charset in our responses from RW and that this occurs when talking with external apis, that define charset in the response?

flowertwig:
@jepp3, correct!

PR #656, opened 2023-12-08
Fix setting loglevel to debug when a request fails due to 404

splitfeed:
Correctly set log level to debug when response status is 404 and ignore any other irrelevant differences in the exceptions.

The old code compared exact instances, but using equals we get the expected behaviour. New test that verifies the behaviour has been added as well.

### DIFF
--- a/jaxrs-api/src/main/java/se/fortnox/reactivewizard/jaxrs/WebException.java
+++ b/jaxrs-api/src/main/java/se/fortnox/reactivewizard/jaxrs/WebException.java
@@ -155,7 +155,7 @@ public class WebException extends RuntimeException {
         if (httpStatus.code() >= 500) {
             return Level.ERROR;
 
-        } else if (httpStatus == HttpResponseStatus.NOT_FOUND) {
+        } else if (httpStatus.equals(HttpResponseStatus.NOT_FOUND)) {
             return Level.DEBUG;
         }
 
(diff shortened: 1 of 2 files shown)


jaxrs-api/src/main/java/se/fortnox/reactivewizard/jaxrs/WebException.java
```diff
@@ -155,7 +155,7 @@ private Level logLevelFromStatus(HttpResponseStatus httpStatus) {
         if (httpStatus.code() >= 500) {
             return Level.ERROR;
 
-        } else if (httpStatus == HttpResponseStatus.NOT_FOUND) {
+        } else if (httpStatus.equals(HttpResponseStatus.NOT_FOUND)) {
```
jepp3, line 158: nice since httpstatus is not an enum. :+1:

jepp3: Have you done some search if we do the same error somewhere else?

jepp3 (approved):

jaxrs-api/src/main/java/se/fortnox/reactivewizard/jaxrs/WebException.java
flowertwig: 🤯 Nice find!

flowertwig (approved):

softqwewasd (approved):

jaxrs-api/src/main/java/se/fortnox/reactivewizard/jaxrs/WebException.java
splitfeed: > Have you done some search if we do the same error somewhere else?

Just checked, nothing found in this repo anyway 👍

PR #664, opened 2023-12-21
Fixing query limit=1 failing when the return type of the DAO method is Mono.

softqwewasd:
It is questionable whether CollectionOptions should even be usable when the return type of a DAO method is Mono. That said, the limit=limit+1 line should only be applied for DAO methods returning Flux.

### DIFF
--- a/dao/src/test/java/se/fortnox/reactivewizard/db/CollectionOptionsTest.java
+++ b/dao/src/test/java/se/fortnox/reactivewizard/db/CollectionOptionsTest.java
@@ -2,6 +2,8 @@ package se.fortnox.reactivewizard.db;
 
 import org.junit.jupiter.api.Test;
 import reactor.core.publisher.Flux;
+import reactor.core.publisher.Mono;
+import reactor.test.StepVerifier;
 import se.fortnox.reactivewizard.CollectionOptions;
 import se.fortnox.reactivewizard.db.config.DatabaseConfig;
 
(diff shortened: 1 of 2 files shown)


dao/src/test/java/se/fortnox/reactivewizard/db/CollectionOptionsTest.java
```diff
@@ -73,12 +75,26 @@ void shouldNotSetLimitAboveMaxLimit() throws SQLException {
...
+        StepVerifier.create(collectionOptionsDao.selectWithDefaultLimit10(collectionOptions))
+            .expectNextCount(10)
+            .verifyComplete();
         mockDb.verifySelect("select * from table LIMIT 11");
         assertThat(collectionOptions.isLastRecord()).isTrue();
     }
 
+    @Test
+    void shouldUseConfiguredDefaultLimitOfOneWithReturnTypeMono() throws SQLException {
+        mockDb.addRows(1);
```
softqwewasd, line 89: I decided to cheat a bit here. The test fails when the num rows > 1 since MockDb does not look at the limit in the query. Having num rows > limit when the return type is Flux works fine because the PagingOperator filters out the remaining rows, but the PagingOperator is never used when working with Monos.

softqwewasd: That said, this has been tested against an actual postgres database and it works as expected.

flowertwig (approved):

PR #678, opened 2024-01-16
"Broken pipe" error handling

nolexa:
`io.netty.channel.unix.Errors.NativeIoException` thrown on a `Broken pipe` error should not be interpreted as an internal server error. It should be treated as one of the conditions that terminate the inbound connection. We cannot refer the `NativeIoException` directly here because it's for internal use only. Instead, we are checking for the parent `IOException` class in a combination with the message text.

### DIFF
--- a/jaxrs/src/main/java/se/fortnox/reactivewizard/ExceptionHandler.java
+++ b/jaxrs/src/main/java/se/fortnox/reactivewizard/ExceptionHandler.java
@@ -73,12 +73,7 @@ public class ExceptionHandler {
             webException = new WebException(BAD_REQUEST, "invalidjson", throwable.getMessage());
         } else if (throwable instanceof WebException we) {
             webException = we;
-        } else if (throwable instanceof ClosedChannelException
-            || throwable instanceof AbortedException
-            || throwable instanceof CancellationException
-            || (throwable instanceof IOException
-            && throwable.getMessage() != null
-            && throwable.getMessage().contains("Broken pipe"))) {
+        } else if (isAnticipatedClientException(throwable)) {
             LOG.atDebug()
                 .setMessage(INBOUND_CONNECTION_CLOSED)
                 .addArgument(request::method)
@@ -100,6 +95,15 @@ public class ExceptionHandler {
             return response.sendString(justOrEmpty(json(webException)));
         }
         return empty();
+    }
+
+    private static boolean isAnticipatedClientException(Throwable throwable) {
+        return throwable instanceof ClosedChannelException
+            || throwable instanceof AbortedException
+            || throwable instanceof CancellationException
+            || (throwable instanceof IOException
+            && throwable.getMessage() != null
+            && throwable.getMessage().contains("Broken pipe"));
     }
 
     private String json(WebException webException) {
--- a/jaxrs/src/main/java/se/fortnox/reactivewizard/ExceptionHandler.java
+++ b/jaxrs/src/main/java/se/fortnox/reactivewizard/ExceptionHandler.java
@@ -73,8 +73,12 @@ public class ExceptionHandler {
             webException = new WebException(BAD_REQUEST, "invalidjson", throwable.getMessage());
         } else if (throwable instanceof WebException we) {
             webException = we;
-        } else if (throwable instanceof ClosedChannelException || throwable instanceof AbortedException || throwable instanceof CancellationException
-            || (throwable instanceof IOException && throwable.getMessage().contains("Broken pipe"))) {
+        } else if (throwable instanceof ClosedChannelException
+            || throwable instanceof AbortedException
+            || throwable instanceof CancellationException
+            || (throwable instanceof IOException
+            && throwable.getMessage() != null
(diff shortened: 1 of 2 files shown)


jaxrs/src/main/java/se/fortnox/reactivewizard/ExceptionHandler.java
```diff
@@ -72,7 +73,8 @@ public Publisher<Void> handleException(HttpServerRequest request, HttpServerResp
             webException = new WebException(BAD_REQUEST, "invalidjson", throwable.getMessage());
         } else if (throwable instanceof WebException we) {
             webException = we;
-        } else if (throwable instanceof ClosedChannelException || throwable instanceof AbortedException || throwable instanceof CancellationException) {
+        } else if (throwable instanceof ClosedChannelException || throwable instanceof AbortedException || throwable instanceof CancellationException
+            || (throwable instanceof IOException && throwable.getMessage().contains("Broken pipe"))) {
```
jepp3, line 77: Could there be a risk of IOException having a null message? It can be done from someone/something throwing an io exception this way?

```
var io = new IOException();

assertThat(io.getMessage())
	.isNull();
```

jaxrs/src/test/java/se/fortnox/reactivewizard/ExceptionHandlerTest.java
```diff
@@ -152,6 +153,22 @@ void shouldLogCancellationExceptionAtDebugLevel() {
         }
     }
 
+    @Test
+    void shouldLogNativeIoExceptionAtDebugLevel() {
```
jepp3, line 157: could you add test with a ioexception thrown with 

a. null message
b. other message

nolexa: Fixed

jaxrs/src/main/java/se/fortnox/reactivewizard/ExceptionHandler.java
```diff
@@ -72,7 +73,8 @@ public Publisher<Void> handleException(HttpServerRequest request, HttpServerResp
             webException = new WebException(BAD_REQUEST, "invalidjson", throwable.getMessage());
         } else if (throwable instanceof WebException we) {
             webException = we;
-        } else if (throwable instanceof ClosedChannelException || throwable instanceof AbortedException || throwable instanceof CancellationException) {
+        } else if (throwable instanceof ClosedChannelException || throwable instanceof AbortedException || throwable instanceof CancellationException
```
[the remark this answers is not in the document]
splitfeed: Sorry for being vague, I meant extracting parts to variables or methods so that the conditions are in a concise and readable format.

Like for example
`if (isExpectedClientException) `

jaxrs/src/main/java/se/fortnox/reactivewizard/ExceptionHandler.java
```diff
@@ -72,7 +73,12 @@ public Publisher<Void> handleException(HttpServerRequest request, HttpServerResp
             webException = new WebException(BAD_REQUEST, "invalidjson", throwable.getMessage());
         } else if (throwable instanceof WebException we) {
             webException = we;
-        } else if (throwable instanceof ClosedChannelException || throwable instanceof AbortedException || throwable instanceof CancellationException) {
+        } else if (throwable instanceof ClosedChannelException
+            || throwable instanceof AbortedException
+            || throwable instanceof CancellationException
+            || (throwable instanceof IOException
```
flowertwig, line 79: I think we should break out checks that are not just `instanceof` into separate methods.

softqwewasd (approved):

jaxrs/src/main/java/se/fortnox/reactivewizard/ExceptionHandler.java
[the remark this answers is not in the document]
nolexa: Yes, that was a good suggestion. Fixed :)

jaxrs/src/main/java/se/fortnox/reactivewizard/ExceptionHandler.java
nolexa: Checks for anticipated exceptions are extracted into a separate method.

jepp3 (approved):

jaxrs/src/main/java/se/fortnox/reactivewizard/ExceptionHandler.java
```diff
@@ -96,6 +97,15 @@ public Publisher<Void> handleException(HttpServerRequest request, HttpServerResp
         return empty();
     }
 
+    private static boolean isAnticipatedClientException(Throwable throwable) {
+        return throwable instanceof ClosedChannelException
+            || throwable instanceof AbortedException
+            || throwable instanceof CancellationException
+            || (throwable instanceof IOException
+            && throwable.getMessage() != null
```
jepp3, line 105: What i think was the idea was to have `|| isBrokenPipe(throwable)` to increase the readability.

splitfeed (approved):

PR #680, opened 2024-01-17
SER-13233 Add support for sorting on multiple columns using CollectinOptions sortby.

YouWillNeverWalkAl1:
Added support for specifying multiple sort by columns in the CollectionOptions sortBy string separated by commas. Each column needs to be defined in allowedSortColumns in order to be used in the sort.
The sortBy string could e.g. contain the string "name asc, id desc" which would result in the corresponding sortby clause.

### DIFF
--- a/dao/src/main/java/se/fortnox/reactivewizard/db/query/parts/CollectionOptionsQueryPart.java
+++ b/dao/src/main/java/se/fortnox/reactivewizard/db/query/parts/CollectionOptionsQueryPart.java
@@ -8,6 +8,10 @@ import se.fortnox.reactivewizard.util.FluxRxConverter;
 
 import java.lang.reflect.Method;
 import java.sql.SQLException;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import static com.google.common.collect.Iterables.indexOf;
 import static java.util.Arrays.asList;
@@ -17,6 +21,7 @@ public class CollectionOptionsQueryPart implements QueryPart {
     private final        int    collectionOptionsArgIndex;
     private final        Query  queryAnnotation;
     private final        boolean isMono;
+    private static final Pattern SORT_BY_PART = Pattern.compile("(?<sortby>[a-zåäö0-9_]+)(\\s+(?<order>asc|desc))?");
 
     public CollectionOptionsQueryPart(Method method) {
         collectionOptionsArgIndex = indexOf(asList(method.getParameterTypes()), CollectionOptions.class::isAssignableFrom);
@@ -36,11 +41,24 @@ public class CollectionOptionsQueryPart implements QueryPart {
             if (collectionOptions != null) {
                 if (collectionOptions.getSortBy() != null) {
                     String sortBy = CamelSnakeConverter.camelToSnake(collectionOptions.getSortBy());
-                    for (String allowed : queryAnnotation.allowedSortColumns()) {
-                        if (allowed.equals(sortBy)) {
-                            collectionOptionsHasOrderBy = true;
-                            addOrderBy(sql, buildOrderBy(collectionOptions.getOrder(), allowed));
-                            break;
+                    var sortByParts = sortBy.split(",");
+                    Collections.reverse(Arrays.asList(sortByParts));
+                    for (var sortByPart: sortByParts) {
+                        Matcher matcher = SORT_BY_PART.matcher(sortByPart);
+                        if (matcher.find()) {
+                            var sortByColumn = matcher.group("sortby");
+                            if (sortByColumn != null && !sortByColumn.isBlank()) {
+                                var order = matcher.group("order");
+                                if (order != null || collectionOptions.getSortBy() != null) {
+                                    for (String allowed : queryAnnotation.allowedSortColumns()) {
+                                        if (allowed.equals(sortByColumn)) {
+                                            collectionOptionsHasOrderBy = true;
+                                            addOrderBy(sql, buildOrderBy(order == null ? collectionOptions.getOrder() : CollectionOptions.SortOrder.valueOf(order.toUpperCase()), allowed));
+                                            break;
+                                        }
+                                    }
+                                }
+                            }
                         }
                     }
                 }
--- a/dao/src/test/java/se/fortnox/reactivewizard/db/CollectionOptionsTest.java
+++ b/dao/src/test/java/se/fortnox/reactivewizard/db/CollectionOptionsTest.java
@@ -171,6 +171,20 @@ class CollectionOptionsTest {
     }
 
     @Test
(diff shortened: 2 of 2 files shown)


JonasHallFnx (changes requested):
Nice feature!

Some things to consider though:
- If there is even a slight risk for sql-injection, I don't think it is worth the risk. I currently see such a risk with this PR.
- Readability of the code is important, especially to assess the risk of the code. I think the readability and structure of the code can be improved.
- Performance is key in code that runs often. If the input looks like the old normal case, you could make an early exit without running regex and other slow stuff. Perhaps queryAnnotation.allowedSortColumns() could be a Set in order to avoid nested loops?

For me the security part is really important. This is one of the few places where an injection could happen. I would like the code to be very easy to read and that the query is only built from strings that are defined by the application, using lookups based on the input, rather than using the input and propagating it to the sql.

YouWillNeverWalkAl1:
I've added that if sortby contains only one column it will continue as before (without using regexp).
As to the security considerations I can only say that if you look at the code you will see that the sort columns inserted comes solely from allowed columns and sort order only from the enum SortOrder (if they try anything else the CollectionOptions.SortOrder.valueOf will fail).
As to the readability of the code there might be improvements to make (as I wrote the code I think it's pretty obvious what it does but for other it might not be). Please feel free to do any changes you see fit but I think this functionality should be included in the framework.

dao/src/main/java/se/fortnox/reactivewizard/db/query/parts/CollectionOptionsQueryPart.java
```diff
@@ -17,6 +21,7 @@ public class CollectionOptionsQueryPart implements QueryPart {
     private final        int    collectionOptionsArgIndex;
     private final        Query  queryAnnotation;
     private final        boolean isMono;
+    private static final Pattern SORT_BY_PART = Pattern.compile("(?<sortby>[a-zåäö0-9_]+)(\\s+(?<order>asc|desc))?");
```
softqwewasd, line 24: Do we need åäö?

dao/src/main/java/se/fortnox/reactivewizard/db/query/parts/CollectionOptionsQueryPart.java
```diff
@@ -36,11 +41,33 @@ public void visit(StringBuilder sql, Object[] args) {
...
+                        for (String allowed : queryAnnotation.allowedSortColumns()) {
+                            if (allowed.equals(sortBy)) {
+                                collectionOptionsHasOrderBy = true;
+                                addOrderBy(sql, buildOrderBy(collectionOptions.getOrder(), allowed));
+                                break;
+                            }
+                        }
+                    } else {
+                        Collections.reverse(Arrays.asList(sortByParts));
+                        for (var sortByPart: sortByParts) {
```
softqwewasd, line 55: Would be nice to filter out the columns that have already been appended to the ordering. For instance, sortBy="name asc, id desc, name asc, id desc, ..." should not become "ORDER BY name ASC, id DESC, name ASC, id DESC, ..." but rather "ORDER BY name ASC, id DESC".

dao/src/main/java/se/fortnox/reactivewizard/db/query/parts/CollectionOptionsQueryPart.java
```diff
@@ -36,11 +41,33 @@ public void visit(StringBuilder sql, Object[] args) {
             if (collectionOptions != null) {
                 if (collectionOptions.getSortBy() != null) {
                     String sortBy = CamelSnakeConverter.camelToSnake(collectionOptions.getSortBy());
-                    for (String allowed : queryAnnotation.allowedSortColumns()) {
-                        if (allowed.equals(sortBy)) {
-                            collectionOptionsHasOrderBy = true;
-                            addOrderBy(sql, buildOrderBy(collectionOptions.getOrder(), allowed));
-                            break;
+                    var sortByParts = sortBy.split(",");
```
softqwewasd, line 44: Could be good to have a check for max length for the query param unless Netty's query decoder already checks that?

dao/src/main/java/se/fortnox/reactivewizard/db/query/parts/CollectionOptionsQueryPart.java
YouWillNeverWalkAl1: I don't know. We are a swedish company...

dao/src/main/java/se/fortnox/reactivewizard/db/query/parts/CollectionOptionsQueryPart.java
YouWillNeverWalkAl1: Would make the code more complicated and would not change the result. It is also a fault made by the user of RW and should not occur.

dao/src/main/java/se/fortnox/reactivewizard/db/query/parts/CollectionOptionsQueryPart.java
YouWillNeverWalkAl1: I do not know. sortBy is however not a new parameter and that change request is not addressing this change.

dao/src/main/java/se/fortnox/reactivewizard/db/query/parts/CollectionOptionsQueryPart.java
softqwewasd: I want to think that åäö is not being used even though Fortnox is a swedish company :). Regardless, it is probably ok to leave it as is.

dao/src/main/java/se/fortnox/reactivewizard/db/query/parts/CollectionOptionsQueryPart.java
softqwewasd: I agree that it would make the code more complicated and would not change the output. I am not sure I understand what you mean that it is a fault made by the user of RW and should not occur. Do you mean that the responsibility of dealing with "sortby=id asc, name desc, id asc, name desc, ..." falls on the application developer?

Even though it would make the code more complicated and doesn't change the output (as well as probably a slight performance impact), I prefer to see the filtering because I find it unreasonable that we would allow an end user to craft a query that looks like that. Furthermore, while it may not be a security concern related to postgres, I have other security concerns that we can discuss in private.

dao/src/main/java/se/fortnox/reactivewizard/db/query/parts/CollectionOptionsQueryPart.java
softqwewasd: Sure, but we are now splitting it and running a regex for each column when sorting by multiple columns compared to only sorting by a single column with an equals comparison? I'd say 256 bytes is a reasonable limit, but we can put it at 1024 to be on the safe side, doesn't make that much of a difference performance wise.

PR #681, opened 2024-01-18
Remove unnecessary reverse lookup

flowertwig:
Remove unnecessary reverse lookup in HttpClient.


client/src/main/java/se/fortnox/reactivewizard/client/RequestBuilder.java
```diff
@@ -73,7 +72,7 @@ public void setServerInfo(InetSocketAddress serverInfo) {
     }
 
     public String getFullUrl() {
-        return serverInfo.getHostName() + ":" + serverInfo.getPort() + uri;
+        return serverInfo.getHostString() + ":" + serverInfo.getPort() + uri;
```
splitfeed, line 75: So this is the actual effective change, right?

flowertwig: Yes. Maybe there should be a comment about not using getHostName. The tests should ensure that, though.

softqwewasd:
Nice!

softqwewasd (approved):

client/src/main/java/se/fortnox/reactivewizard/client/RequestBuilder.java
jepp3: @flowertwig have you verified that it works? By looking at where the InetSocketAddress serverInfo variable is constructed:
https://github.com/FortnoxAB/reactive-wizard/blob/master/client/src/main/java/se/fortnox/reactivewizard/client/HttpClient.java#L130
it looks like the InetSocketAddress will do DNS to IP lookup in the constructor. It cannot be right that it will resolve DNS to IP here. The HTTP client should get the URL that was configured in this case example.com. Otherwise, the IP can never change or am I wrong? 

```java

    /**
     *
     * Creates a socket address from a hostname and a port number.
     * <p>
     * An attempt will be made to resolve the hostname into an InetAddress.
     * If that attempt fails, the address will be flagged as <I>unresolved</I>.
     * <p>
     * If there is a security manager, its {@code checkConnect} method
     * is called with the host name as its argument to check the permission
     * to resolve it. This could result in a SecurityException.
     * <P>
     * A valid port value is between 0 and 65535.
     * A port number of {@code zero} will let the system pick up an
     * ephemeral port in a {@code bind} operation.
     *
     * @param   hostname the Host name
     * @param   port    The port number
     * @throws IllegalArgumentException if the port parameter is outside the range
     * of valid port values, or if the hostname parameter is {@code null}.
     * @throws SecurityException if a security manager is present and
     *                           permission to resolve the host name is
     *                           denied.
     * @see     #isUnresolved()
     */
    public InetSocketAddress(String hostname, int port) {
        checkHost(hostname);
        InetAddress addr = null;
        String host = null;
        try {
            addr = InetAddress.getByName(hostname);
        } catch(UnknownHostException e) {
            host = hostname;
        }
        holder = new InetSocketAddressHolder(host, addr, checkPort(port));
    }
```

flowertwig: @jepp3 I've verified that It "works" in some cases at least, but you make a good point. It's true that the host name will be discarded if the lookup is successful when creating the InetSocketAddress. That behaviour, however, was not introduced in this PR.

jepp3: We need to resolve this issue before it happens. There is a lot of communication with external partners, that are affected by this flaw. I am thinking of connecting to banks, email providers, and other smaller companies, and their APIs.  I'm fine with doing it in another issue. But I would prefer it to be resolved it the same issue as it is highly connected to the same area of domain and code. :+1:

jepp3 (approved):

noxreview:
Looks good! :+1: :1st_place_medal:

PR #688, opened 2024-01-29
Check minimum rows affected when executing statement in a batch

pavlo-liapota:
If statement is executed in a transaction then `monoSink` and `fluxSink` are not set for a statement.

If statement is also batched with another statement, then `batch()` and `batchExecuted()` methods are called instead of `execute()` method.

Current `batchExecuted()` method does nothing as `fluxSink` and `monoSink` are always `null`, so `ensureMinimumReached()` method is never called.


flowertwig (changes requested):
Thank you for this nice find! Some tests would be nice, though.

flowertwig (approved):

softqwewasd (approved):

PR #698, opened 2024-02-20
PLAT-3447: Log real JaxRs resource class names

nolexa:
This update will try to resolve the real JaxRs resource class names and log those names instead of the proxies at the startup and initialization time.


jepp3 (approved):
I approve it. but it would really be nice to have test case to see how the output would look like

nolexa:
That's how the new logging output looks now:
```
INFO  [2024-02-20 21:07:35.595] se.fortnox.reactivewizard.jaxrs.JaxRsResources: 
    GET	/health (com.fortnox.reactivewizard.health.HealthResource.checkHealth)
    ...
```

nolexa:
> I approve it. but it would really be nice to have test case to see how the output would look like

I'll see if I can construct a test case for it. Basically, the output looks exactly same way as it does for non-proxied classes.

For instance, this is an internal `reactivewizard` endpoint, that is not proxied:
```
GET	/memory (com.fortnox.reactivewizard.admin.memory.MemoryResource.memoryInfo)
```
With this update, the external endpoints are presented in the same manner, with the real class names. It's been tested on Fortnox applications.

PR #702, opened 2024-02-27
Use bundled JDK and enable Maven caching

splitfeed:
Enabled Maven caches
https://github.com/actions/setup-java?tab=readme-ov-file#caching-maven-dependencies
This should save a bunch of time from the builds and, if working properly, not introducing a higher risk of issues :)

Updated YAML formatting in editorconfig and cleaned up the format of the .github/ files

Added myself to dependabot and removed flowertwig 

Updated deploy_on_release.yml to make it work with drafts as well, since `created` only triggers when releases are directly created instead of made as drafts and then published

### DIFF
--- a/.github/workflows/deploy_on_release.yml
+++ b/.github/workflows/deploy_on_release.yml
@@ -2,7 +2,7 @@ name: Automatic deploy on release
 run-name: Deploy of ${{ github.ref }} on release
 on:
   release:
-    types: [ created ]
+    types: [ published ]
 
 jobs:
   get_tag:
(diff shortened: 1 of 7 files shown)


.github/workflows/deploy_on_release.yml
```diff
@@ -1,25 +1,26 @@
 name: Automatic deploy on release
 run-name: Deploy of ${{ github.ref }} on release
 on:
-    release:
-        types: [ created ]
+  release:
+    types: [ published ]
```
softqwewasd, line 5: Nice!

softqwewasd (approved):

PR #706, opened 2024-03-04
DAO module clean up and refactoring

pavlo-liapota:
**1. Remove resultConverter from ReactiveStatementFactory**
`ReactiveStatementFactory` always creates `Mono` if dao method return type is `Mono` and `Flux` if return type is `Flux`. So it looks like there is no need for a converter.
The check that return type is Mono or Flux was moved to `ReactiveStatementFactory`.

**2. Remove paramSerializer from DbProxy**
There are no usages of `paramSerializer` field in `DbProxy`. I have checked internal RW too. So it looks like it is safe to remove it.

**3. Move dao method related code from ReactiveStatementFactory to DaoMethodHandler**
Currently `DbProxy` has two "roles":
- proxies any dao method
- has fields related to query execution (connectionScheduler, databaseConfig)

And `ReactiveStatementFactory` has two "roles":
- builds Mono or Flux for query executions
- handles dao method execution and stores precalculated information for a specific dao method

My idea is to distribute "roles" in the following way:
- DbProxy: proxies any dao method
- ReactiveStatementFactory: executes the query
- DaoMethodHandler: handles dao method execution and stores precalculated information for a specific dao method

So this refactoring has moved the following code:
- precalculated info for dao method and dao method handling from `ReactiveStatementFactory` to a new class `DaoMethodHandler`
- everything needed for query execution from `DbProxy` to `ReactiveStatementFactory`

**Note:** previously if query execution was slow then the query sql was logged, now the metric name related to the dao method will be logged instead.
We did this change so that we don't need to pass sql text as a parameter.
And it will be easier for a developer to find this query in the code.

So previously log message may look like this:
```
Slow query: select * from table where name=:name
time: 6000
```
Now it may look like this:
```
Slow query: DAO_type:query_method:com.fortnox.reactivewizard.dao.MyDao.doSomethingMono_1 time: 6000
```

### DIFF
--- a/dao/src/main/java/se/fortnox/reactivewizard/db/DbProxy.java
+++ b/dao/src/main/java/se/fortnox/reactivewizard/db/DbProxy.java
@@ -6,14 +6,11 @@ import reactor.core.scheduler.Scheduler;
 import reactor.core.scheduler.Schedulers;
 import se.fortnox.reactivewizard.db.config.DatabaseConfig;
 import se.fortnox.reactivewizard.db.paging.PagingOutput;
-import se.fortnox.reactivewizard.db.statement.DbStatementFactory;
 import se.fortnox.reactivewizard.db.statement.DbStatementFactoryFactory;
-import se.fortnox.reactivewizard.db.transactions.ConnectionScheduler;
 import se.fortnox.reactivewizard.metrics.Metrics;
 import se.fortnox.reactivewizard.util.DebugUtil;
 import se.fortnox.reactivewizard.util.ReflectionUtil;
 
-import javax.annotation.Nullable;
 import java.lang.reflect.InvocationHandler;
 import java.lang.reflect.Method;
 import java.lang.reflect.Proxy;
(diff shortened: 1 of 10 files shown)


mkfortnox:
And just to confirm that we're on the same page here this PR does not add support for execution of arbitrary SQL text unless you implement the full Statement interface etc, right?

sj-robert:
Suggestion: reword from "Slow query:" to "Slow execution"; most often it's not the database that's slow, it's the thread being starved due to garbage collection.

splitfeed:
I would really like to see a much higher grade of code coverage here. Regardless of coverage of the old code, it was battle tested and known to work. If we are to make these changes and feel confident that it works, good coverage of tests is key.

Most of the coverage from ReactiveStatementFactoryTest comes from a single test that focuses only on releasing schedule workers, so I doubt much of the coverage is actually real coverage.

dao/src/main/java/se/fortnox/reactivewizard/db/ReactiveStatementFactory.java
```diff
@@ -92,46 +90,54 @@ private Mono<Object> getResultMono(StatementContext statementContext) {
...
-                resultFlux = resultFlux.onErrorResume(thrown -> {
-                    queryFailure.initCause(thrown);
-                    return Flux.error(queryFailure);
-                });
-            }
-            resultFlux = pagingOutput.apply(resultFlux, args);
-            resultFlux = Flux.from(metrics.measure(resultFlux, this::logSlowQuery));
-            resultFlux = resultFlux.onBackpressureBuffer(RECORD_BUFFER_SIZE);
-            return decorated(resultConverter.apply(resultFlux), statementContext);
+    public <T> Mono<T> createMono(
```
splitfeed, line 93: The old create method actually had javadocs that you removed, would you please add them to your new create-methods?

softqwewasd:
Have you checked that these changes work with the internal RW? Is there a branch in the internal RW showing this?

dao/src/main/java/se/fortnox/reactivewizard/db/DbProxy.java
```diff
@@ -1,97 +1,57 @@
...
-    private final Scheduler scheduler;
-    protected final Map<Method, ReactiveStatementFactory> statementFactories;
-    private final ConnectionScheduler connectionScheduler;
-    protected final Function<Object[], String> paramSerializer;
-    private final DatabaseConfig databaseConfig;
+    protected final Map<Method, DaoMethodHandler> handlers;
+    private final ReactiveStatementFactory reactiveStatementFactory;
 
     @Inject
-    public DbProxy(DatabaseConfig databaseConfig,
```
splitfeed, line 44: I'm perplexed about the removal of the JsonSerializerFactory, was it never used anywhere?

splitfeed (changes requested):
I cannot say much about the functionality of the dao module, but I have a few concerns quality wise. Hopefully it should not be too painful to get some decent code coverage and add javadocs :)

pavlo-liapota:
> Have you checked that these changes work with the internal RW? Is there a branch in the internal RW that we can view?

Yes. I have created a branch `feature/NOJIRA-update-RW-dependency`.

dao/src/main/java/se/fortnox/reactivewizard/db/ReactiveStatementFactory.java
```diff
@@ -92,46 +90,54 @@ private Mono<Object> getResultMono(StatementContext statementContext) {
...
+        Metrics metrics,
+        Supplier<Statement> statementSupplier
+    ) {
+        var statementContext = new StatementContext(statementSupplier, connectionScheduler);
+        Mono<T> resultMono = getResultMono(statementContext);
+        if (shouldAddDebugErrorHandling()) {
+            resultMono = resultMono.onErrorResume(thrown ->
+                Mono.error(new RuntimeException(QUERY_FAILED, thrown))
```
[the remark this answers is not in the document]
pavlo-liapota: > Will this not result in creating new RuntimeExceptions for each error, as opposed as the old code reusing the same one?

In the old code, each time method `ReactiveStatementFactory#create` is called, the new instance of exception is created.
The only difference is that now exception is created only if error happens while previously it was created even if error didn't happened.

The reusability of the exception instance could only happen in the old code if lambda provided to method `onErrorResume` is called several times. I suppose this may happen if we add a retry. Is there any benefit in reusing the exception instance in this case?

pavlo-liapota:
> Have you checked if any other dependents other than internal RW are impacted by these changes?

No, but I don't see an issue if others are dependent too. They would need to update the constructor usages as was done in internal RW.

pavlo-liapota:
> I would really like to see a much higher grade of code coverage here. Regardless of coverage of the old code, it was battle tested and known to work. If we are to make these changes and feel confident that it works, good coverage of tests is key.
> 
> Most of the coverage from ~ReactiveStatementFactoryTest~DaoMethodHandlerTest comes from a single test that focuses only on releasing schedule workers, so I doubt much of the coverage is actually real coverage.
> 
> Edit: I did not notice that ReactiveStatementFactoryTest was renamed to DaoMethodHandlerTest. This just means that DaoMethodHandlerTest is poorly covered and ReactiveStatementFactory doesn't have a single unit test.

I have renamed `DaoMethodHandlerTest` back to `ReactiveStatementFactoryTest` and have created a separate unit test for `DaoMethodHandler`.
`ReactiveStatementFactory` is tested through `DbProxyTest`.

pavlo-liapota:
> Hopefully it should not be too painful to get some decent code coverage and add javadocs :)

I have added javadocs and more tests.
To increase code coverage more, I would need to set `DebugUtil.IS_DEBUG` to true. Do you know how to do it in the unit test?

softqwewasd (approved):

jepp3:
I looked through the pr and i just have some general thoughts that I do not think you need to resolve now to get the pr merged. 

* The class DaoMethodHandler was a bit hard to understand. I would have placed the variables in a record and extracted the create method to a method in the Proxy class. ( if the caching is needed for other things than the reactivestatementfactory. Like this: 

```
record DaoMethodContext(Method method,
                               DbStatementFactory statementFactory,
                               PagingOutput pagingOutput,
                               Metrics metrics) { }
```
                               
Then the proxy class would look something like this: 

```
 @Override
    public Object invoke(Object proxy, Method method, Object[] methodArgs) throws Throwable {
        DaoMethodContext methodContext = methodContexts.get(method);
        if (methodContext == null || DebugUtil.IS_DEBUG) {
            if (DebugUtil.IS_DEBUG) {
                // Need to get the actual interface method in order to get updated annotations
                method = ReflectionUtil.getRedefinedMethod(method);
            }

            methodContext = new DaoMethodContext(
                method,
                dbStatementFactoryFactory.createStatementFactory(method),
                new PagingOutput(method),
                createMetrics(method)
            );
            methodContexts.put(method, methodContext);
        }

        return getExecutableStatement(methodArgs, methodContexts.get(method));
    }

    public static Publisher<Object> getExecutableStatement( Object[] args, DaoMethodContext daoMethodContext) {
        this.reactiveStatementFactory.createFlux() // the code from the create method.
    }
```

  * It could be good to rename the args parameter to methodArgs to be more clear, where it is used. 
  * the ifs and else are a bit verbose in the daoMethodHandler class. you could maybe make it a bit more clear by writing it like this : 
    
```
      Class<?> returnType = daoMethodContext.method().getReturnType();
        
        if(returnType.isAssignableFrom()) {
            
        }
        if(returnType.isAssignableFrom()) {
            
        }
        throw 
```
    
    
  But overall it looks nice, hope it works well for you :)

jepp3 (approved):

pavlo-liapota:
@splitfeed Can you review again?

splitfeed (approved):

PR #748, opened 2024-05-06
Re-add old constructor to avoid breaking changes for downstream

splitfeed:
No reason to cause issues for the services that call DbProxy, but marking the constructor as deprecated to indicate that it needs changing.


softqwewasdfortnox (approved):

softqwewasd (approved):

dao/src/main/java/se/fortnox/reactivewizard/db/DbProxy.java
```diff
@@ -54,6 +55,20 @@ protected DbProxy(ReactiveStatementFactory reactiveStatementFactory,
         this.handlers = handlers;
     }
 
+    /**
+     * @deprecated Use {@link #DbProxy(DatabaseConfig, ConnectionProvider)} instead
```
nolexa, line 59: I wonder why the replacement link is referring the two arguments constructor instead of the one that takes all three arguments omitting only `JsonSerializerFactory`? I mean this constructor: 
```
DbProxy(DatabaseConfig databaseConfig, ConnectionProvider connectionProvider,
        DbStatementFactoryFactory dbStatementFactoryFactory)
```

nolexa (approved):

PR #750, opened 2024-05-10
PLAT-3553: Flushing the logs at the server shutdown

nolexa:
When RW server is shut down, the latest portions of the logs entries are not flushed to the storage, and are often lost. This update enables a new handler that will be called at the end of the shutdown procedure in order to flush the logs.

### DIFF
--- a/logging-api/pom.xml
+++ b/logging-api/pom.xml
@@ -1,0 +1,12 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <artifactId>reactivewizard-logging-api</artifactId>
+
+    <parent>
+        <groupId>se.fortnox.reactivewizard</groupId>
+        <artifactId>reactivewizard-parent</artifactId>
+        <version>999.9.9-SNAPSHOT</version>
+    </parent>
+
+</project>
--- a/server/src/main/java/se/fortnox/reactivewizard/server/RwServer.java
+++ b/server/src/main/java/se/fortnox/reactivewizard/server/RwServer.java
@@ -13,6 +13,7 @@ import reactor.netty.http.server.HttpServer;
 import reactor.netty.http.server.HttpServerRequest;
 import reactor.netty.http.server.HttpServerResponse;
 import se.fortnox.reactivewizard.RequestHandler;
+import se.fortnox.reactivewizard.logging.LoggingShutdownHandler;
 import se.fortnox.reactivewizard.server.modifiers.NoContentFixConfigurer;
 import se.fortnox.reactivewizard.server.modifiers.RequestSizesConfigurer;
 
@@ -47,28 +48,32 @@ public class RwServer extends Thread {
     private final ServerConfig config;
     private final ConnectionCounter connectionCounter;
     private final DisposableServer server;
+    private final LoggingShutdownHandler loggingShutdownHandler;
     private static Runnable blockShutdownUntil;
 
     @Inject
     public RwServer(ServerConfig config, CompositeRequestHandler compositeRequestHandler, ConnectionCounter connectionCounter,
-        Set<ReactorServerConfigurer> serverConfigurers) {
-        this(config, connectionCounter, createHttpServer(config, serverConfigurers), compositeRequestHandler);
+        Set<ReactorServerConfigurer> serverConfigurers, LoggingShutdownHandler loggingShutdownHandler) {
+        this(config, connectionCounter, createHttpServer(config, serverConfigurers), compositeRequestHandler, loggingShutdownHandler);
     }
 
-    public RwServer(ServerConfig config, CompositeRequestHandler compositeRequestHandler, ConnectionCounter connectionCounter) {
-        this(config, compositeRequestHandler, connectionCounter, Set.of(new NoContentFixConfigurer(), new RequestSizesConfigurer(config)));
+    public RwServer(ServerConfig config, CompositeRequestHandler compositeRequestHandler,
+                    ConnectionCounter connectionCounter, LoggingShutdownHandler loggingShutdownHandler) {
+        this(config, compositeRequestHandler, connectionCounter,
+            Set.of(new NoContentFixConfigurer(), new RequestSizesConfigurer(config)), loggingShutdownHandler);
     }
 
     RwServer(ServerConfig config, ConnectionCounter connectionCounter, HttpServer httpServer,
-             CompositeRequestHandler compositeRequestHandler) {
-        this(config, connectionCounter, httpServer, compositeRequestHandler, null);
+             CompositeRequestHandler compositeRequestHandler, LoggingShutdownHandler loggingShutdownHandler) {
+        this(config, connectionCounter, httpServer, compositeRequestHandler, null, loggingShutdownHandler);
     }
 
     RwServer(ServerConfig config, ConnectionCounter connectionCounter, HttpServer httpServer, CompositeRequestHandler compositeRequestHandler,
-        DisposableServer disposableServer) {
+        DisposableServer disposableServer, LoggingShutdownHandler loggingShutdownHandler) {
         super("RwServerMain");
         this.config = config;
         this.connectionCounter = connectionCounter;
+        this.loggingShutdownHandler = loggingShutdownHandler;
 
         if (config.isEnabled()) {
             if (disposableServer != null) {
@@ -141,7 +146,8 @@ public class RwServer extends Thread {
     }
 
     void registerShutdownHook() {
-        Runtime.getRuntime().addShutdownHook(new Thread(() -> shutdownHook(config, server, connectionCounter)));
+        Runtime.getRuntime().addShutdownHook(new Thread(() ->
+            shutdownHook(config, server, connectionCounter, loggingShutdownHandler)));
     }
 
     /**
@@ -155,7 +161,8 @@ public class RwServer extends Thread {
         RwServer.blockShutdownUntil = blockShutdownUntil;
     }
 
-    static void shutdownHook(ServerConfig config, DisposableServer server, ConnectionCounter connectionCounter) {
+    static void shutdownHook(ServerConfig config, DisposableServer server, ConnectionCounter connectionCounter,
+                             LoggingShutdownHandler loggingShutdownHandler) {
         LOG.info("Shutdown requested. Waiting {} seconds before commencing.", config.getShutdownDelaySeconds());
         try {
             Thread.sleep(config.getShutdownDelaySeconds() * 1000);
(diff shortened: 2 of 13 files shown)


server/src/main/java/se/fortnox/reactivewizard/server/RwServer.java
```diff
@@ -193,6 +210,25 @@ static void awaitShutdownDependency(int shutdownTimeoutSeconds) {
         LOG.info("Shutdown dependency completed, continue...");
     }
 
+    static void awaitLoggingShutdown(int shutdownTimeoutSeconds) {
+        if (loggingShutdownRunner == null) {
+            return;
+        }
+
+        LOG.info("Wait for logging to shut down");
```
jepp3, line 218: Mabe its even easier to express the wait with reactor?

	Mono.fromRunnable(runnableStuff)
			.timeout(Duration.ofSeconds(3))
			.subscribe();

server/src/test/java/se/fortnox/reactivewizard/server/RwServerTest.java
```diff
@@ -204,6 +205,21 @@ void shouldAwaitShutdownDependency() {
         ));
     }
 
+    @Test
+    void shouldAwaitShutdownLogging() {
```
jepp3, line 209: No test for the timeout logic?

server/src/main/java/se/fortnox/reactivewizard/server/RwServer.java
```diff
@@ -155,6 +158,18 @@ public static void registerShutdownDependency(Runnable blockShutdownUntil) {
         RwServer.blockShutdownUntil = blockShutdownUntil;
```
JonasHallFnx, line 158: Perhaps it would be possible to change this into a list of Runnable so you can use it instead and avoid duplicated code? I wrote this code some years ago and the only reason for not having a list was hat I didn't need it at that time.

server/src/main/java/se/fortnox/reactivewizard/server/RwServer.java
```diff
@@ -155,6 +158,18 @@ public static void registerShutdownDependency(Runnable blockShutdownUntil) {
...
     }
 
+    /**
+     * Registers a Runnable to be run at shutdown of RW server in order to flush and shut down the logging.
+     *
+     * @param loggingShutdownRunner the runnable
+     */
+    public static void registerLoggingShutdown(Runnable loggingShutdownRunner) {
+        if (RwServer.loggingShutdownRunner != null && loggingShutdownRunner != null) {
```
splitfeed, line 167: Is your intention that running registerLoggingShutdown(null) means that you unregister a previous hook?
I would rather expect this method to do nothing or throw an error if it was passed null, so this confuses me a bit :)

server/src/test/java/se/fortnox/reactivewizard/server/RwServerTest.java
splitfeed, line 209: Could we also add one test verifying the handling of a shutdown timeout?

server/src/main/java/se/fortnox/reactivewizard/server/RwServer.java
```diff
@@ -44,10 +44,13 @@ public class RwServer extends Thread {
         "application/json"
     ));
 
+    private static final int LOGGING_SHUTDOWN_TIMEOUT_SEC = 3;
```
splitfeed, line 47: Is there a particular reasoning behind using 3 seconds? I would expect it to be much faster, but might it also be worth waiting a few more extra seconds to get the last logs in case it is a bit slow?

softqwewasd, line 47: Nice changes overall, I think it is a good start! Perhaps the timeout config should be configurable in the ServerConfig?

nolexa: No specific reason for just 3 seconds. I anticipate that logging generally be pretty quick, but we still need a timeout. The shutdown timeout of 20 seconds is too long. 3 seconds should probably be enough in 99.99% cases. But of course, we can make it whatever figure we like.

nolexa: This is a local technical timeout. It does not bear any semantics that would be meaningful for any other stages of the shutdown process. It just guards of extremely unlikely conditions when the logging shutdown does not respond. I thought this parameter would obstruct the ServerConfig with something that is not interesting in the global context. Maybe I'm mistaken. We can add it to the config if you think it's necessary.

server/src/main/java/se/fortnox/reactivewizard/server/RwServer.java
nolexa: That was my first intention as well, but then I realised that we're not applying the handlers universally at one single point in the shutdown procedure. If it was like that, a list of Runnable would be a natural response. Instead, we attach handlers to the hooks at different stages. So waiting for dependencies is one stage, shutting down the logging is a different stage. It happens at the very last step of the procedure. So the Runnable handlers should be identifiable in order to call them where they are needed.

server/src/main/java/se/fortnox/reactivewizard/server/RwServer.java
softqwewasd: I want to avoid changing this in the open source RW later on, but I do agree that 3 seconds probably covers all cases so there should be no reason to change this in the future. I somewhat agree that ServerConfig is probably wrong to use here, perhaps just a system property would otherwise work. Nevertheless, no changes needed :)

server/src/main/java/se/fortnox/reactivewizard/server/RwServer.java
nolexa: It was not exactly my intention. Actually, I copied this registration method from the dependency shutdown hook. I did not check all the possible outcomes. But yes, you're right, calling the method with null argument will in effect unregister the handler. I'd rather agree that this does not make sense. If we do not allow to reassign a handler why should it be allowed to unassign one? That does not sound consistent.

server/src/test/java/se/fortnox/reactivewizard/server/RwServerTest.java
nolexa: Good idea :)

server/src/main/java/se/fortnox/reactivewizard/server/RwServer.java
nolexa: It looks nice when expressed via reactor operations. But I suspect that the initial motivation behind using the bare thread semantics was that we want simple low level means of control during the shutdown procedure. I guess that's not the best moment to rely on the reactor's thread behaviour.

server/src/test/java/se/fortnox/reactivewizard/server/RwServerTest.java
nolexa: Will do 👍

softqwewasd (approved):

server/src/test/java/se/fortnox/reactivewizard/server/RwServerTest.java
splitfeed: This and @jepp3:s comment above is resolved by not having a timeout anymore, right?

splitfeed (changes requested):

nolexa:
There's a new approach implemented. Previous solution required that the local RW project would directly reach the reactivewizard-server module in order to register a hook handler. That appeared to be a bad idea. The new approach decouples the dependencies through exposing a `LoggingShutdownHandler` interface, letting the local project to implement the interface. This puts an additional requirement on the binders to make the new interface available for injection.

logging-api/pom.xml
```diff
@@ -0,0 +1,12 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <artifactId>reactivewizard-logging-api</artifactId>
```
nolexa, line 4: The module has been renamed to `reactivewizard-loggin-api`

server/src/main/java/se/fortnox/reactivewizard/server/RwServer.java
nolexa: There's no timeout anymore, so this code is gone.

server/src/test/java/se/fortnox/reactivewizard/server/RwServerTest.java
nolexa: The timeout was removed. Nothing to test.

server/src/main/java/se/fortnox/reactivewizard/server/RwServer.java
nolexa: The approach has changed. We're not using Runnable for this purpose. We inject an interface instead.

server/src/test/java/se/fortnox/reactivewizard/server/RwServerTest.java
nolexa: This is not relevant any longer. The timeout has been removed.

server/src/main/java/se/fortnox/reactivewizard/server/RwServer.java
nolexa: The timeout has been removed. The context is outdated.

server/src/main/java/se/fortnox/reactivewizard/server/RwServer.java
```diff
@@ -193,6 +202,18 @@ static void awaitShutdownDependency(int shutdownTimeoutSeconds) {
         LOG.info("Shutdown dependency completed, continue...");
     }
 
+    static void executeLoggingShutdown(LoggingShutdownHandler loggingShutdownHandler) {
```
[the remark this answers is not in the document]
nolexa: Sure, why not? I followed the pattern of the other shutdown hook - dependencies shutdown. Maybe that one also deserves to be extracted..

server/src/main/java/se/fortnox/reactivewizard/server/RwServer.java
nolexa: There's no timeout anymore. This constant is removed.

server/src/main/java/se/fortnox/reactivewizard/server/RwServer.java
nolexa: This code has been removed. There's no handler registration done this way.

server/src/main/java/se/fortnox/reactivewizard/server/RwServer.java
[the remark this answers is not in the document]
nolexa: The problem with RwServer being dependent on LoggingShutdownHandler is that `shutdownHook` is a static method. It cannot use the injected class members; it needs all dependencies to be provided as the method arguments. So even if we delegate some of the hook functions to other classes, the entry point to the chain of static calls will remain in `RwServer.class`, hence the dependencies.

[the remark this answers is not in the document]
nolexa: Refactoring of the shutdown-related stuff is spared for a separate PR. It is not to be done here.

jepp3 (approved):

splitfeed (approved):

PR #754, opened 2024-05-16
PLAT-13791: Setting a database connection timezone via DatabaseConfig.

nolexa:
If a timezone is not passed via database connection properties, Postgres JDBC driver uses the system default timezone. This can lead to flaky behaviours if the default timezone is not yet initialised at the time when connection is created. Instead, we added a timezone property to DatabaseConfig in order to explicitly specify a timezone when creating a connection.

### DIFF
--- a/dbmigrate/src/main/java/se/fortnox/reactivewizard/dbmigrate/LiquibaseMigrate.java
+++ b/dbmigrate/src/main/java/se/fortnox/reactivewizard/dbmigrate/LiquibaseMigrate.java
@@ -1,5 +1,6 @@
 package se.fortnox.reactivewizard.dbmigrate;
 
+import com.google.common.base.Strings;
 import jakarta.inject.Inject;
 import liquibase.CatalogAndSchema;
 import liquibase.Contexts;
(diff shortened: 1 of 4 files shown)


softqwewasd:
Have you tested these changes together with the changes in internal RW to confirm that this works?

nolexa:
> Have you tested these changes together with the changes in internal RW to confirm that this works?

The property-based approach did not work. The latest update with "SET TIME ZONE" initialisation does work all way through. I tested it in three tiers: oos RW, internal RW, and an internal application, both running a server and BDD tests.

dbmigrate/src/main/java/se/fortnox/reactivewizard/dbmigrate/LiquibaseMigrate.java
```diff
@@ -37,11 +40,20 @@
...
+    private static final Logger LOG = LoggerFactory.getLogger(LiquibaseMigrate.class);
+
     private final List<Liquibase> liquibaseList;
 
     @Inject
     public LiquibaseMigrate(LiquibaseConfig liquibaseConfig) throws LiquibaseException, IOException {
         JdbcConnection conn = new JdbcConnection(getConnection(liquibaseConfig));
+        if (!Strings.isNullOrEmpty(liquibaseConfig.getTimeZone())) {
+            try {
+                conn.createStatement().execute("SET TIME ZONE '%s'".formatted(liquibaseConfig.getTimeZone()));
```
jepp3, line 52: Does this approach not require us to always need to keep in mind to set the time zone in every single place we create a new connection?:thinking:

PR #755, opened 2024-05-20
Downgrading jackson due to concurrency issues

softqwewasd:
https://github.com/FasterXML/jackson-databind/pull/4430


softqwewasd:
> Are we expecting any regressions from downgrading this? I remember seeing something in regards to Kotlin that was fixed by this version, if I’m not mistaken?

The Kotlin bug was introduced in Jackson-databind 2.17.0 and was resolved in 2.17.1. We _should_ be fine downgrading to 2.16.2 but I will double check this before merging.

TommyPersson (approved):

PR #769, opened 2024-06-17
Allowing blocking calls inside Jackson's DeserializerCache::_createAndCacheValueDeserializer resolves the IllegalMonitorStateException

softqwewasd:
Allowing blocking calls inside Jackson's DeserializerCache::_createAndCacheValueDeserializer resolves the IllegalMonitorStateException.

See https://github.com/FasterXML/jackson-databind/pull/4430

This PR resolves the issue we had when upgrading to Jackson 2.17.1. The upgrade to 2.17.1 caused some of our tests to fail with IllegalMonitorStateException. By configuring Blockhound to allow blocking calls inside _createAndCacheValueDeserializer, IllegalMonitorStateException is no longer thrown. You can read more about this issue here https://github.com/FasterXML/jackson-databind/pull/4430#issuecomment-2150633048.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -79,7 +79,7 @@
 
         <reactor-bom.version>2023.0.6</reactor-bom.version>
         <netty.version>4.1.110.Final</netty.version>
-        <jackson-bom.version>2.16.2</jackson-bom.version>
+        <jackson-bom.version>2.17.1</jackson-bom.version>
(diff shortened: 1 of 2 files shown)


TommyPersson (approved):

softqwewasd:
> So this is a dependency upgrade and fix of a test?

Pretty much. I've updated the PR description with some additional context.

PR #797, opened 2024-09-19
PLAT-3546: Validating Config-files

nolexa:
Validation of the config files seems to be broken for a long time. `ValidatingConfigFactory` did not have effect because its binding was always overridden by the default config factory from the bootstrap binding.

For some reason bootstrap binding was giving the highest priority, thus overriding the bindings of all the auto-bind modules. I could not find any justification for this behaviour. The only class that is bound in the bootstrap module is `ConfigFactory`, and we want it to be overridden by `ValidatingConfigFactory`. It looks like it was a bug.

The modules precedence is fixed so that the bootstrap is at the lowest place, overridden by the rest of the modules.

### DIFF
--- a/bootstrap/src/main/java/se/fortnox/reactivewizard/Main.java
+++ b/bootstrap/src/main/java/se/fortnox/reactivewizard/Main.java
@@ -7,9 +7,6 @@ import com.google.inject.name.Names;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import se.fortnox.reactivewizard.binding.AutoBindModules;
-import se.fortnox.reactivewizard.config.ConfigFactory;
-
-import java.nio.file.NoSuchFileException;
 
 /**
  * Main application entry point. Will scan for all modules on the classpath and run them.
@@ -27,41 +24,16 @@ public class Main {
      */
     public static void main(String[] args) {
         try {
-            ConfigFactory  configFactory  = createConfigFactory(args);
             Module bootstrap = new AbstractModule() {
                 @Override
                 protected void configure() {
                     bind(String[].class).annotatedWith(Names.named("args")).toInstance(args);
-                    bind(ConfigFactory.class).toInstance(configFactory);
                 }
             };
             Guice.createInjector(new AutoBindModules(bootstrap));
         } catch (Exception exception) {
             LOG.error("Caught exception at startup.", exception);
             System.exit(-1);
-        }
-    }
-
-    /**
-     * Prepares a ConfigFactory before setting up Guice.
-     * <p>
-     * As logging can be part of the configuration file and the configuration file
-     * could be missing, we have a side effect of setting up and initializing a LoggingFactory
-     * that doesn't depend on ConfigFactory, if the configuration file is missing.
-     *
-     * @param args commandline arguments
-     * @return A ConfigFactory based on a configuration file.
-     * @throws NoSuchFileException If the configuration file cannot be found.
-     */
-    private static ConfigFactory createConfigFactory(String[] args) throws NoSuchFileException {
-        try {
-            return new ConfigFactory(args);
-        } catch (RuntimeException runtimeException) {
-            Throwable cause = runtimeException.getCause();
-            if (cause != null && cause.getClass().isAssignableFrom(NoSuchFileException.class)) {
-                throw (NoSuchFileException)cause;
-            }
-            throw runtimeException;
         }
     }
 }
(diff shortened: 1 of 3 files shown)


jepp3 (approved):

binding/src/main/java/se/fortnox/reactivewizard/binding/AutoBindModules.java
```diff
@@ -84,20 +84,13 @@ public void configure(Binder binder) {
...
-            if (mergedModule == null) {
-                mergedModule = m;
-            } else {
-                mergedModule = Modules.override(mergedModule).with(m);
-            }
+            mergedModule = Modules.override(mergedModule).with(m);
             m.preBind();
         }
 
-        mergedModule = Modules.override(mergedModule)
```
jepp3, line 98: checked if the test injector uses this to defectively override modules?

jepp3 (changes requested):
failing build

validation/src/main/java/se/fortnox/reactivewizard/validation/ValidationModule.java
```diff
@@ -56,4 +54,10 @@ private <T> void bindValidationProxy(Binder binder, Provider<ValidatorUtil> vali
     private <T> Provider<T> validatingProvider(Class<T> iface, Provider<T> wrappedProvider, Provider<ValidatorUtil> validatorUtilProvider) {
         return () -> ValidatingProxy.create(iface, wrappedProvider.get(), validatorUtilProvider.get());
     }
+
+    @Override
+    public Integer getPrio() {
+        // We want this module to override the default binding of ConfigFactory
+        return 110;
```
splitfeed, line 61: Is there any way of creating a test to validate that this actually works, and more importantly continues to work in the future as well?

binding/src/test/java/se/fortnox/reactivewizard/binding/AutoBindModulesTest.java
```diff
@@ -28,12 +28,12 @@ void shouldBindImplementationsToTheirInterfaces() {
     }
 
     @Test
-    void shouldUseBootStrapBindingOverDefault() {
```
splitfeed, line 31: This seems to be a very conscious choice, did you find any hints on why it was this way? It could help prevent reintroducing bugs that this might have resolved.

splitfeed (changes requested):

binding/src/test/java/se/fortnox/reactivewizard/binding/AutoBindModulesTest.java
nolexa: No, I don't have any clue why it's done this way. The bootstrap module binds only two things:
``` java
                    bind(String[].class).annotatedWith(Names.named("args")).toInstance(args);
                    bind(ConfigFactory.class).toInstance(configFactory);
``` 
So it's either the default ConfigFactory class or whatever ConfigFactory compatible class defined in the override modules. I could not find hints why the default factory should always override.

binding/src/main/java/se/fortnox/reactivewizard/binding/AutoBindModules.java
nolexa: TestInjector should indeed override the modules. I made a special case for it.

binding/src/test/java/se/fortnox/reactivewizard/binding/AutoBindModulesTest.java
nolexa: Maybe this order of overrides was made just for a single purpose: to promote the TestInjector in DbProxyTest. When this case is fixed, both tests and applications work fine.

validation/src/main/java/se/fortnox/reactivewizard/validation/ValidationModule.java
nolexa: I added a test that compares priorities between this module and the config module.

bootstrap/src/main/java/se/fortnox/reactivewizard/Main.java
```diff
@@ -41,27 +36,4 @@ protected void configure() {
             System.exit(-1);
         }
     }
-
-    /**
-     * Prepares a ConfigFactory before setting up Guice.
-     * <p>
-     * As logging can be part of the configuration file and the configuration file
-     * could be missing, we have a side effect of setting up and initializing a LoggingFactory
-     * that doesn't depend on ConfigFactory, if the configuration file is missing.
```
nolexa, line 50: It is not clear if this justification is still vald.

jepp3: it is no longer valid.

jepp3 (approved):

magnusp (approved):

splitfeed (approved):

